PR #2, opened 2019-05-14
Solution by Vlad Polets

vpolets:


src/style.css
```diff
@@ -0,0 +1,72 @@
...
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf) format("truetype");
+  font-weight: 800;
+  font-style: normal;
+}
+
+* {
+  box-sizing: border-box;
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
```
gestap: No need to use * for `font-family` and `font-size`
Add it to body.

PR #3, opened 2019-05-14
solution

Laliquenoare:


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
+  border-radius: 3px;
+  background: url(./images/Search.svg) no-repeat 26px center, linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  font-weight: bold;
+}
+
+.search-form__input_small:focus {
+  border: 1px solid #e1e7ed;
+  font-weight: bold;
```
gestap: `f-w` must be changed only for input text, not for placeholder.

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  outline: none;
+  color: #3d4e61;
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+}
+
+.search-form__input_big {
+  font-size: 16px;
+  line-height: 22px;
+  padding: 25px 0 23px 62px;
```
gestap: why did you add this padding?
you need only left

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  background: url(./images/Search.svg) no-repeat 26px center;
+  background-size: 19px 19px;
+  height: 70px;
+}
+
+.search-form__input_small {
+  height: 42px;
+  font-size: 14px;
+  line-height: 16px;
+  padding: 12px 0 11px 33px;
```
gestap: !

PR #6, opened 2019-05-14
Search bars AirBnB: big, small (Volkovnytskyi, Kurochka)

ekaterinakur:


src/style.css
```diff
@@ -0,0 +1,92 @@
...
+form .search-bar:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+form .search-bar:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+  box-shadow: none;
+  font-weight: 600;
```
gestap: `f-w` must be changed only for input text, not for placeholder.

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+  top: 15px;
+  left: 13px;
+}
+
+form .search-bar {
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  width: 100%;
+  height: 100%;
```
gestap: it's not quite good.
If you add more components to form, your input will be not what you expect.
input height should not depend on the height of the form.

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+  position: absolute;
+  z-index: 1;
+  top: 25px;
+  left: 26px;
+}
+
+form.small-search::before {
+  content: "";
+  background-image: url(./images/Search.svg);
+  background-size: 11px 11px;
```
gestap: no need duplicated size, use `cover` or `contain`

PR #63, opened 2019-09-03
Fs aug19 kostia s

kostya-flern:


src/index.html
```diff
@@ -13,7 +13,20 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
```
gestap: create form

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+  top: 15px;
+}
+
+.big-search-bar:hover,
+.small-search-bar:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.big-search-bar:focus,
+.small-search-bar:focus {
```
gestap: font-weight must change

src/style.css
```diff
@@ -0,0 +1,73 @@
+@font-face {
```
gestap: you need to use this style.
Now you only connect them to this document

PR #141, opened 2020-02-24
Solution

crash2bash:
- [DEMO LINK](https://crash2bash.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://crash2bash.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,73 @@
...
+
+.search__big-wrapper {
+  position: relative;
+  margin: 20px 0;
+}
+
+.search__big-wrapper::before {
+  content: "";
+  position: absolute;
+  background-image: url(images/Search.svg);
```
earthhuman41: These styles are repeated in small, so i think you need to make default class .search and specify common styles for it and apply it for big search bar and then you can add "search search--small" for small search to specify specific styles for it

earthhuman41: These styles are repeated in small search. I think you don't need "search__big" you need just "search__small" all styles for big make as default for search bar in "search__blocks", then in small you can override desired properties adding to "search-blocks" "search-blocks--small". So small search will has 2 classes. And rename common class "search__..." just "search" to form and input can be search__field or etc

crash2bash: Хмм, и вправду это будет более элегантным решением. Объявить стили, а после переопределить их в схожем, но слегка модифицированном блоке. Спасибо!

PR #142, opened 2020-02-25
initial commit

coastin:
- [DEMO LINK](https://coastin>.github.io/layout_search-bar-airbnb/)
    - [TEST REPORT LINK](https://coastin.github.io/layout_search-bar-airbnb/report/html_report/)


vkryvytskyy:
Вставляй ссылки на демо и тест репорт без пробелов вначале, иначе они становятся не кликабельными

vkryvytskyy:
Ссылки не работают

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  height: 19px;
+  margin-top: 25px;
+  margin-left: 26px;
+}
+  
+.small-search {
+  height: 42px;
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 16px;
```

PR #145, opened 2020-02-25
solution

hheight:
- [DEMO LINK](https://hheight.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://hheight.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,58 @@
...
+}
+
+.search-bar--size-small {
+  font-size: 14px;
+  line-height: 19px;
+  height: 42px;
+  padding-left: 33px;
+  background-position: top 14px left 12px;
+  background-size: 11px 11px;
+}
```

PR #146, opened 2020-02-26
Solution

SeiAlek:
- [DEMO LINK](https://SeiAlek.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://SeiAlek.github.io/layout_search-bar-airbnb/report/html_report/)

What wrong with font  Avenir on my solution?


src/style.css
```diff
@@ -0,0 +1,75 @@
...
+  line-height: 22px;
+  height: 70px;
+}
+
+.searchbar--small .search-input--wrapper::before {
+  height: 11px;
+  width: 11px;
+  margin-top: 15px;
+  margin-left: 13px;
+}
```

PR #147, opened 2020-02-26
Solution

Qoa11a:
[DEMO LINK](https://Qoa11a.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://Qoa11a.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  width: 100%;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
+  box-shadow: rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  margin-bottom: 20px;
+}
+
+.searchbar-large {
```
earthhuman41: i think you dont need this class, just make .searchbar as default and apply to it default sizes, and then make .searchbar-small and override necessary styles

PR #149, opened 2020-02-26
solution

veli4kodima:
- [DEMO LINK](https://veli4kodima.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://veli4kodima.github.io/layout_search-bar-airbnb/report/html_report/)


src/index.html
```diff
@@ -9,6 +9,24 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-  <h1>Search bar airbnb</h1>
+
+<div class="wrapper">
+  <div class="search-box search-box-big">
+    <form action="" data-qa="big">
+      <input type="image" src="images/Search.svg" class="icon-search icon-search-big">
```
vkryvytskyy: `<input type="image" >` causes page to reload, try to use image as **background** into `<input type="text">`

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  line-height: 22px;
+  font-size: 16px;
+
+}
+
+.search-box-small input[type="search"] {
+  padding-top: 10px;
+  padding-bottom: 11px;
+  padding-left: 33px;
+  /* Text */
```
vkryvytskyy: remove comments

src/index.html
```diff
@@ -9,6 +9,24 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-  <h1>Search bar airbnb</h1>
+
+<div class="wrapper">
+  <div class="search-box search-box-big">
+    <form action="" data-qa="big">
+      <input type="text" class="icon-search icon-search-big">
+      <input type="search" placeholder="Try &quot;Los Angeles&quot;">
```
vkryvytskyy: remove first input, and add background-image prop to second

src/style.css
```diff
@@ -0,0 +1,69 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+* {
+  padding: 0;
+  font-family: Avenir, serif;
+}
```
vkryvytskyy: Try to avoid selector *. Yes in these tasks it doesn't influence performance, but in bigger projects, it can load browser

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+  background-size: 19px 19px;
+}
+
+.search-box-big input[type="search"] {
+  padding-top: 23px;
+  padding-bottom: 23px;
+  padding-left: 62px;
+  line-height: 22px;
+  font-size: 16px;
+}
```
vkryvytskyy: move it tot search-box
We don't have any basic search bar, so why we should use there big and small search class names? We have search-input and search-input - small (--small). Therefore, we set all styles that are supposed to be for the "large" input for all inputs in general, and with the modifier small we change only the necessary values to make our input smaller.

src/style.css
```diff
@@ -0,0 +1,66 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+body * {
+  padding: 0;
+  font-family: Avenir, serif;
+}
```
vkryvytskyy: Try to avoid selector *. Yes in these tasks it doesn't influence performance, but in bigger projects, it can load browser

vkryvytskyy: apply font directly to input, not to all inside body

vkryvytskyy: Try to avoid increasing the specificity of the selector. One class is enough to identify the elements on the page. No need to write something like .block.colored-block. Use .block and .block--colored instead, and apply each of them to your element

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+
+.search-box {
+  position: relative;
+  width: 100%;
+}
+
+.search-box-big {
+  margin: 20px 0;
+}
```

src/index.html
```diff
@@ -9,6 +9,22 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-  <h1>Search bar airbnb</h1>
+
+<div class="wrapper">
+  <div class="search-box search-box-big">
```
vkryvytskyy: remove 
```suggestion
  <div class="search-box">
```

src/style.css
```diff
@@ -0,0 +1,67 @@
...
+  padding: 0;
+}
+
+.search-box {
+  position: relative;
+  width: 100%;
+  margin: 20px 0;
+}
+
+.search-box input[type="search"] {
```
vkryvytskyy: add class to input, instead of using input[type="search"] selector

src/style.css
```diff
@@ -61,7 +61,7 @@ body {
   background-size: 11px 11px;
 }
 
-.search-box--small input[type="search"]:focus {
+.search-box--small .search-box-field {
```

src/style.css
```diff
@@ -51,7 +51,7 @@ body {
   margin: 0;
 }
 
-.search-box--small input[type="search"] {
+.search-box--small .search-box-field {
```

src/style.css
veli4kodima: я наверное тупой(

veli4kodima:
ЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕЕ!!!!!!!!!!!!!!!!!!!!!!

PR #150, opened 2020-02-26
Develop

Ishchuk-R:
[DEMO LINK](https://Ishchuk-R.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://Ishchuk-R.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,49 @@
...
+.input-search:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.input-search:focus {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  font-weight: 500;
+}
+
+.input-lg {
```

Ishchuk-R: fixed it

earthhuman41:
Clear icon too close to right border, make gap for it
Screen Shot 2020-02-27 at 11 11 19 AM

Ishchuk-R:
> Clear icon too close to right border, make gap for it
> Screen Shot 2020-02-27 at 11 11 19 AM

Ishchuk-R:
done

PR #152, opened 2020-02-26
Develop

omelchenkom:
- [DEMO LINK](https://omelchenkom.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://omelchenkom.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,54 @@
...
+  background-repeat: no-repeat;
+  background-position: 25px 24px;
+  background-image: url(images/Search.svg);
+  color: #3d4e61;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  box-sizing: border-box;
+}
+
+.search-bar.small {
```

src/style.css
```diff
@@ -0,0 +1,54 @@
...
+  background-image: url(images/Search.svg) linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  background-repeat: no-repeat;
+  background-position: 25px 24px;
+  background-size: 19px;
+  font-weight: bold;
+  border-radius: 3px;
+  outline: none;
+}
+
+.search-bar.small:focus {
```
vkryvytskyy: here also

omelchenkom: в следующий раз можете писать сразу все ошибки

omelchenkom: в следующий раз можете писать сразу все ошибки

PR #154, opened 2020-02-26
solution

OlySpring1:
- [DEMO LINK](https://olyspring1.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://olyspring1.github.io/layout_search-bar-airbnb/report/html_report/)

Я знаю, що не пройшла тести, але мені потрібна допомога. Що саме в мене неправильно.

### DIFF
--- a/src/font/font.css
+++ b/src/font/font.css
@@ -1,0 +1,6 @@
+
+@font-face {
+  font-family: "Avenir";
+  src: url("./fonts/Avenir-Book.ttf");
+  font-weight: 300;
+}
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,67 @@
+input {
+  width: 100%;
+  font-family: Avenir, sans-serif;
+}
+
+.icon {
+  background: url(images/Search.svg);
+  background-repeat: no-repeat;
+}
+
+.icon-big {
+  background-position: 26px 25px;
+  background-size: 19px 19px;
+}
+
+.icon-small {
+  background-position: 13px 15px;
+  background-size: 11px 11px;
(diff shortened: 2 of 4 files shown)


src/font/font.css
```diff
@@ -0,0 +1,6 @@
+
+@font-face {
+  font-family: "Avenir";
+  src: url("./fonts/Avenir-Book.ttf");
+  font-weight: 300;
+}
```
vkryvytskyy: where did you include this file into your page?

src/style.css
```diff
@@ -0,0 +1,67 @@
...
+
+.icon-big {
+  background-position: 26px 25px;
+  background-size: 19px 19px;
+}
+
+.icon-small {
+  background-position: 13px 15px;
+  background-size: 11px 11px;
+}
```

src/style.css
```diff
@@ -0,0 +1,67 @@
...
+  margin-top: 40px;
+  padding: 12px 226px 10px 33px;
+  font-size: 14px;
+  line-height: 19px;
+  color: #3d4e61;
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+}
```

src/style.css
```diff
@@ -0,0 +1,55 @@
...
+  fill: #687c94;
+}
+
+.search-input:focus {
+  border-radius: 3px;
+  font-weight: 900;
+  fill: #687c94;
+}
+
+.search-input.search-input-small {
```

PR #157, opened 2020-02-26
I tried

panasiuk-yuriy:
- [DEMO LINK](https://panasiuk-yuriy.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://panasiuk-yuriy.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,56 @@
...
+
+.small-search-bar {
+  height: 42px;
+}
+
+.big-block,
+.small-block {
+  position: relative;
+  margin-top: 20px;
+}
```

PR #162, opened 2020-02-27
'solution_develop1'

FaychakA:
- [DEMO LINK](https://FaychakA.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://FaychakA.github.io/layout_search-bar-airbnb/report/html_report/)


vkryvytskyy:
image
it looks strange in firefox

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+
+.small_input {
+  height: 42px;
+  position: relative;
+  background-position: top 12px left 14px;
+  background-size: 11px 11px;
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 19px;
+}
```

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  font-size: 16px;
+  line-height: 22px;
+  color: #3d4e61;
+}
+
+.small_input::placeholder {
+  font-size: 14px;
+  line-height: 19px;
+  color: #3d4e61;
+}
```
vkryvytskyy: you could add styles directly to input, not its placeholder

vkryvytskyi (changes requested):

8hokage (approved):

PR #166, opened 2020-02-27
solution

YaroslavMakarov:
- [DEMO LINK](https://YarolavMakarov.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://YaroslavMakarov.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,15 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-  <h1>Search bar airbnb</h1>
+<form class="label">
+  <label for="big-search" data-qa="keypress">
+    <input id="big-search" class="search-big search" type="search" placeholder="Try 'Los Angeles'" data-qa="big">
+  </label>
+</form>
+<form>
+  <label for="small-search" >
+    <input id="small-search" class="search-small search" type="search" placeholder="Try 'Los Angeles'" data-qa="small">
+  </label>
+</form>
 </body>
 </html>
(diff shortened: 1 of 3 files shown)


YaroslavMakarov:
Не вийшло нормально підключити шрифт

src/style.css
```diff
@@ -0,0 +1,50 @@
...
+
+.search-small {
+  margin-top: 20px;
+  height: 42px;
+  background-position: 12px 14px;
+  padding-left: 33px;
+  background-size: 11px 11px;
+  font-size: 14px;
+  line-height: 19px;
+}
```

src/index.html
```diff
@@ -9,6 +9,15 @@
...
 <body>
-  <h1>Search bar airbnb</h1>
+<form class="label">
+  <label for="big-search" data-qa="keypress">
+    <input id="big-search" class="search-big search" type="search" placeholder="Try 'Los Angeles'" data-qa="big">
+  </label>
+</form>
+<form>
+  <label for="small-search" >
+    <input id="small-search" class="search-small search" type="search" placeholder="Try 'Los Angeles'" data-qa="small">
```
vkryvytskyy: split string longer than 80 symbols to few lines by attributes like this:
```
<input 
  id="small-search" 
  class="search-small search" 
  type="search"
...
```

src/style.css
```diff
@@ -0,0 +1,49 @@
...
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search-input:focus {
+  border-radius: 3px;
+  outline: none;
+  font-weight: 900;
+}
+
+.search-input__big {
```

PR #174, opened 2020-02-27
Solution

Vladimir-Beliy:
- [DEMO LINK](https://Vladimir-Beliy.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Vladimir-Beliy.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  position: absolute;
+  display: block;
+  width: 19px;
+  height: 19px;
+  background: url(images/Search.svg);
+  left: 26px;
+  top: 25px;
+}
+
+.search-big {
```

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  border-radius: 3px;
+}
+
+/* --------------- search-small --------------- */
+
+.search-form--smal {
+  position: relative;
+}
+
+.search-form--smal::before {
```

earthhuman41: You dont need repeat same styles for small and big

Vladimir-Beliy: ooo, i got it)

PR #176, opened 2020-02-27
5 tests failed. What to do with the font?

DrozhzhynY:
- [DEMO LINK](https://DrozhzhynY.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://DrozhzhynY.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  background-image: url(images/Search.svg) ;
+  position: absolute;
+  width: 19px;
+  height: 19px;
+  top: 25px;
+  left: 26px;
+}
+
+.big-search__input:hover,
+.small-search__input:hover {
```

DrozhzhynY: Changed classes

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+}
+
+.search {
+  display: block;
+  margin: 20px 0;
+  position: relative;
+}
+
+.search__input,
+.search__input-small {
```

DrozhzhynY: fixed. Code not repeat.

earthhuman41:
You dont need make placeholder bold on focus just input text

DrozhzhynY:
fixed placeholder

PR #180, opened 2020-02-27
Solution v0.1

ShapovalDenys:
https://ShapovalDenys.github.io/layout_search-bar-airbnb/

https://ShapovalDenys.github.io/layout_search-bar-airbnb/report/html_report/


src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  box-sizing: border-box;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  font-size: 16px;
+  line-height: 22px;
+  color: rgb(61, 78,91);
+  margin: 20px 0;
+  background: url("./images/Search.svg") no-repeat;
+  background-position: 25px 26px;
+}
```
vkryvytskyy:
```
.small_search_form_input {
  height: 42px;
  width: 100%;
  padding-left: 33px;
  display: block;
  border: 1px solid rgb(255, 231,237);
  box-sizing: border-box;
  border-radius: 4px;
  font-size: 14px;
  line-height: 19px;
  font-weight: 300;
  color: rgb(61, 78,91);
  margin: 20px 0;
  background: url("./images/Search.svg") no-repeat;
  background-position: 13px 15px;
  background-size: 11px;
}
```
Чем отличаются эти два класса, кроме размера?

vkryvytskyy: We don't have any basic search bar, so why we should use there big and small search class names? We have search-input and search-input - small (--small). Therefore, we set all styles that are supposed to be for the "large" input for all inputs in general, and with the modifier small we change only the necessary values to make our input smaller.

PR #183, opened 2020-02-27
sollution

OlenaZubashenko:
https://olenazubashenko.github.io/layout_search-bar-airbnb/


src/style.css
```diff
@@ -0,0 +1,55 @@
...
+  color: #3d4e61;
+  background: #fff url(./images/Search.svg) no-repeat;
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  outline: none;
+}
+
+.search__panel--big {
```

OlenaZubashenko: done

PR #198, opened 2020-04-22
Solution

kvalenty:
- [DEMO LINK](https://kvalenty.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://kvalenty.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+.input-small {
+  padding-left: 33px;
+  height: 42px;
+  font-family: Avenir;
+  font-size: 14px;
+  line-height: 19px;
+  color: #3d4e61;
+}
```
irenhh, line 82: consider moving repeated styles to some general class

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+body,
+.input-font {
+  font-family: "Avenir";
+  font-weight: 300;
+  color: #3d4e61;
+}
```
irenhh, line 11: do body and input-font need the same styles to be written explicitly?

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  width: 100%;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+}
+
+.search-bar__input:hover {
+  background: #fff;
```
irenhh, line 34: if you have some styles on the element, there is no need to write them again for hover state or some other state

PR #211, opened 2020-04-24
Solution

Kussmich:
- [DEMO LINK](https://kussmich.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://kussmich.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,69 @@
...
+  font-family: "Avenir";
+  font-weight: 300;
+
+  color: #3d4e61;
+
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+
+  background-image: url(./images/Search.svg);
```
irenhh, line 31: you need to add image with pseude-elements or some img with position absolute

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+.search-item::placeholder {
+  font-weight: 300;
+}
+
+.search-item:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search-item:focus {
+  box-shadow: none;
```
irenhh, line 54: there is some shadow on focus according to the mockup

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+  font-weight: 900;
```
irenhh, line 55: inputted text is always the same, it doesn't change on focus

PR #230, opened 2020-07-01
Solution

1nsider21:
- [DEMO LINK](https://1nsider21.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://1nsider21.github.io/layout_search-bar-airbnb/report/html_report/)


src/index.html
```diff
@@ -1,14 +1,43 @@
...
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Search bar airbnb</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="big-search-bar" data-qa="big">
+      <img
+        src="images/Search.svg"
+        alt="search"
+        class="big-img img"
```
irenhh, line 18: try to avoid duplicating tag names in class names

irenhh: how about `search-icon`

src/index.html
```diff
@@ -1,14 +1,43 @@
...
+      >
+      <input
+        type="search"
+        name="search"
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+        class="big-input-search search"
```
irenhh, line 25: try with `search-bar` instead of `search` and `input-big` instead of that long name

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+body {
+  font-family: "Avenir", sans-serif;
+}
+
+.big-search-bar {
+  margin: 20px 0;
+}
+
+.img {
+  position: absolute;
```
irenhh, line 19: position is a property, which is not recommended to move separately since when you want to recheck your code and you try to understand why your elements behave like so, it will be hard to find out where this property is written

just leave it on those elements you need to position absolutely

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+  height: 19px;
+  margin-left: 26px;
+  margin-top: 25px;
+}
+
+.small-img {
+  width: 11px;
+  height: 11px;
+  margin-left: 13px;
+  margin-top: 15px;
```
irenhh, line 33: try with shorthand property

1nsider21:
I fixed all the comments. I don't know why you can't see

PR #234, opened 2020-07-01
solution

tykhan:
- [DEMO LINK](https://tykhan.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://tykhan.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,69 @@
...
+  padding-left: 62px;
+}
+
+.smaller-bar {
+  font-size: 14px;
+  height: 42px;
+  padding-left: 33px;
+}
+
+.big-dandruff {
```
irenhh, line 35: big what? :)

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+.small-dandruff {
+  left: 13px;
+  top: 15px;
+  height: 11px;
+  width: 11px;
+}
+
+.icon {
+  position: absolute;
+}
```
irenhh, line 51: position is a property, which is good to have separately for each element, no need to move it to general class

src/style.css
tykhan: ahah, лупа in english, for native speakers will be clear:)

tykhan: but anyway, I'll change it)

PR #238, opened 2020-07-02
Solution

1nsider21:
- [DEMO LINK](https://1nsider21.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://1nsider21.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,72 @@
...
+}
+
+.container {
+  position: relative;
+  margin: 20px 0;
+}
+
+.search-icon {
+  position: absolute;
+}
```
irenhh, line 21: position is a property that is better to be placed to each element separately

PR #252, opened 2020-07-02
solution with comment

Andrii256:
- [DEMO LINK](https://andrii256.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://andrii256.github.io/layout_search-bar-airbnb/report/html_report/)

**ATTENTION**
I have a qustion (answer please in slack. I wont be notified)
Why my big input have suitable placeholder color without ::placeholder styling.
 But my small input has black color without this? (I can`t to find why)
(deteils are on style.css 78line. And there are not th same statement of .big_input, but big input have right color of placeholder)

+ I have remark
Delete please text shadow in figma maket. https://prnt.sc/tajrxw

+add to the task, that vertical margin should be 20px


src/style.css
```diff
@@ -0,0 +1,78 @@
...
+.big_container::before {
+  content: "";
+  display: block;
+  position: absolute;
+  top: 25.5px;
+  left: 26px;
+  z-index: 1;
+  width: 19px;
+  height: 19px;
+  background-image: url("/images/Search.svg");
```
irenhh, line 48: https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md checklist

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  color: #3d4e61;
+}
+
+.big_container::before {
+  content: "";
+  display: block;
+  position: absolute;
+  top: 25.5px;
+  left: 26px;
+  z-index: 1;
```
irenhh, line 45: do you really need this?

Andrii256: yes https://prnt.sc/takjf3

src/style.css
Andrii256: soluted, chek again please

Andrii256: *inputs dont inherit font-family from body

src/style.css
Andrii256: I changed ::before on <image> element. So that i need not z-index anymore, but I need position: absolute.
Without this image will not cover part of input.

And with position relative there will remain space of image in its started position. I think absolute position is the best way

src/style.css
Andrii256: "font family is usually written for body tag and inherited by others"
input dont inherit font family from body. They are exception. And there are not need to add font family for usual elements like p or div

everything other are fixed

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+  font-weight: 900;
+}
+
+.search-container {
+  position: relative;
+  margin: 20px 0;
+}
+
+.search-image {
+  position: absolute;
```
irenhh, line 23: don't move position out of other styles for current element, it's ok to repeat it

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.search-input {
+  width: 100%;
+  box-sizing: border-box;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  font-family: Avenir, Arial, sans-serif;
```
irenhh, line 32: write this in body => add inherit for inputs

PR #267, opened 2020-07-03
Solution

b0hdnan:
- [DEMO LINK](https://b0hdnan.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://b0hdnan.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  box-sizing: border-box;
+}
+
+.container {
+  position: relative;
+  margin: 20px auto;
+}
+
+.search-icon {
+  position: absolute;
```
irenhh, line 26: don't move position out of other styles, it's ok to repeat writing it

b0hdnan: Thanks. I'll take it into account.

PR #333, opened 2020-11-08
Search bars

tamara-ostapets:
- [DEMO LINK](https://tamara-ostapets.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://tamara-ostapets.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,76 @@
+* {
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url("fonts/Avenir-Book.ttf");
+}
+
+body {
+  font-family: Avenir, sans-serif;
+  /* font-weight: 300; */
+  color: #3d4e61;
+}
+
+.form {
+  position: relative;
+  margin-top: 20px;
+}
+
+.search {
+  font-family: Avenir, sans-serif;
+  width: 100%;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url("fonts/Avenir-Book.ttf");
+}
+
+body {
+  font-family: Avenir, sans-serif;
+  /* font-weight: 300; */
```
vkryvytskyy, line 12: remove comments

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+.small > .icon {
+  position: absolute;
+  left: 13px;
+  top: 15px;
+  width: 11px;
+  height: 11px;
+  display: inline-block;
+}
```
vkryvytskyy, line 76: don't increase selectors specificity, remove `>`. You can deal with it without `>`.

vkryvytskyy, line 76: you should have form, and form--large/form--small. But not large and small form at the same time. One form - is default, another - is modified(smaller than default one, ore bigger).

vkryvytskyy (changes requested):

PR #349, opened 2020-11-11
solution (search bar airbnb)

5Mountains:
- [x] [DEMO LINK](https://5mountains.github.io/layout_search-bar-airbnb/)
- [x] [TEST REPORT LINK](https://5mountains.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,87 @@
+@font-face {
+  font-family: Avenir;
+  src: url("/fonts/Avenir.ttc"),
+    url("/fonts/Avenir-Book.ttf"),
+    url("/fonts/Avenir-Heavy.ttf");
+}
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+}
+
+.wrapper {
+  position: relative;
+  margin-top: 20px;
+}
+
+.input {
+  width: 100%;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  background: #fff;
+  border-radius: 4px;
+  font-weight: 900;
+  color: #3d4e61;
+  font-family: inherit;
+}
+
+.input::placeholder {
+  font-weight: 300;
+}
+
+.big__input {
+  height: 70px;
+  padding-left: 62px;
+  font-size: 16px;
+  line-height: 22px;
+}
+
+.small__input {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
+@font-face {
+  font-family: Avenir;
+  src: url("/fonts/Avenir.ttc"),
+    url("/fonts/Avenir-Book.ttf"),
+    url("/fonts/Avenir-Heavy.ttf");
+}
```
vkryvytskyy, line 6: https://stackoverflow.com/a/49010440

vkryvytskyy (approved):

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  font-size: 16px;
+  line-height: 22px;
+}
+
+.small__input {
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 19px;
+  height: 42px;
+}
```
vkryvytskyy, line 48: you should have default search-bar and big/small. Not big and small at the same time. One of them is just search-bar without big/small word.

vkryvytskyy (changes requested):

src/style.css
5Mountains: Am, I took this approach because I found these fonts in the project folder, so I decided to use them all. So, in the future, I should add fonts via links in the head? Thanks!

5Mountains:
Changes were done

src/style.css
5Mountains: Could you please explain to me about this stuff? 
In this case, I should create another CSS file only with fonts and put it before CSS main style as well I should implement @import this fonts CSS file to the main CSS?)

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  left: 26px;
+  top: 25px;
+}
+
+.wrapper__small::before {
+  width: 11px;
+  height: 11px;
+  left: 13px;
+  top: 15px;
+}
```
vkryvytskyy, line 71: same stuff like with inputs, default wrapper and big/small.
--small is modifier, if you are using BEM

vkryvytskyy (approved):

src/style.css
5Mountains: I'm a newbie with the using BEM *(  
I'll change it. Thanks!

PR #362, opened 2020-11-12
added two search bars

SofiiaAsieieva:


src/style.css
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.search-input:focus {
+  background: rgba( 0,0,0,0);
+  font-weight: 900;
+}
+
+.search-input::placeholder {
+  font-weight: normal;
+}
```
vsm2k17, line 105: placeholder does not inherit font from parents elements

vsm2k17 (changes requested):
Screenshot 2020-11-12 at 16 52 12
these files shouldn't be in your PR

vsm2k17 (approved):

PR #475, opened 2021-03-25
solution

KapToH337:
[DEMO LINK](https://KapToH337.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://KapToH337.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,30 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="form form-1" data-qa="big">
+      <img src="images\Search.svg"
+           alt="Search-img"
+           class="big__img img__search"
+      >
+      <input
+        type="search"
+        id="big_search"
+        class="big__search search"
+        placeholder="Try &#34Los Angeles&#34"
+      >
+    </form>
+
+    <form class="form form-2" data-qa="small">
+      <img src="images\Search.svg"
+           alt="Search-img"
+           class="small__img img__search"
+      >
+      <input
+        type="search"
+        id="small_search"
+        class="small__search search"
+        placeholder="Try &#34Los Angeles&#34"
+      >
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,88 @@
+@font-face {
+  font-family: Avenir-Book;
+  src: url(fonts\Avenir-Book.ttf);
+}
+@font-face {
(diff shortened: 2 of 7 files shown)


src/index.html
```diff
@@ -11,6 +11,31 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="form form-1" data-qa="big">
+      <img src="images\Search.svg"
+           alt="Search-img"
+           class="big__img img__search"
+      >
```
VinnieJ-2k20, line 18: Fix indentations.

```suggestion
      <img
        src="images\Search.svg"
        alt="Search-img"
        class="big__img img__search"
      >
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  color: #3d4e61;
+}
+
+.big__search:hover {
+  border: 1px solid #3d4e61;
+}
+
+.big__search:focus {
+  font-family: Avenir-Heavy, Arial, Helvetica, sans-serif;
+}
```
VinnieJ-2k20, line 57: You have same styles here for big__search and small__search.
Create and apply common class.

VinnieJ-2k20 (changes requested):
You are on the right way.
Details in comments.

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+
+.small__search:hover,
+.big__search:hover {
+  border: 1px solid #3d4e61;
+}
+
+.small__search:focus,
+.big__search:focus {
+  font-family: Avenir-Heavy, Arial, Helvetica, sans-serif;
+}
```
VinnieJ-2k20, line 79: Again, you think in the right direction, but let's make it even more simple.
You'll get to know more about this issue when you'll learn BEM methodology next week.

Let's select some base class for search input.
You already have `search` class that you don't style at all, by the way.
Our big input will be some base option.
Into this class we'll add all the styles we need for big input (incl. :hover, ::placeholder etc.)
We'll apply this class to big input in HTML (`<input class="search" ... `)

Then we have option for small input.
We'll apply two classes here - base class and class with modifier: (`<input class="search search--small" ...`)
And into `search--small` class in CSS we'll add all styles that small input should have, but only those, that differ from big search. So we won't have to write `position: absolute;` or `display: block;` for the second time.

VinnieJ-2k20 (changes requested):
Almost done.
Wrote some comments.
Ask, if you have additional questions.

VinnieJ-2k20 (approved):

PR #486, opened 2021-03-26
Solution

herkoss:
https://herkoss.github.io/layout_search-bar-airbnb/
https://herkoss.github.io/layout_search-bar-airbnb/report/html_report/


vsm2k17 (approved):
good job

PR #498, opened 2021-03-26
created airbnb

YuriiLiso:


src/style.css
```diff
@@ -0,0 +1,66 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
```
varseniuk, line 3: you should specify `font-weight` as well here

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+
+.icon_small_search_bar {
+  position: absolute;
+  width: 11px;
+  height: 11px;
+  padding: 15px 9px 16px 13px;
+}
+
+.input_small_bar,
+.input_big_bar {
```
varseniuk, line 36: You may also use two classes - one which specifies common styles for both big and small inputs, and two other, that specifies exact styles for two inputs separately (e.g:
`class="searchInput searchInput_small"
class="searchInput searchInput_large"`

varseniuk (changes requested):
Let's make some small fixes

varseniuk (approved):
👍 
But read my comments about fontFace to use it in future correctly

src/style.css
```diff
@@ -0,0 +1,68 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: AvenirBold;
```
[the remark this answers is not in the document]
YuriiLiso: Ok, thanks. I understood.

PR #513, opened 2021-03-28
search bar

BoikoEva:
- [DEMO LINK](https://<boikoeva>.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://<boikoeva>.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,76 @@
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: bold;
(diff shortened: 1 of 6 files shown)


vsm2k17 (changes requested):
demo link and test link does not work :(

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+.form {
+  position: relative;
+  display: flex;
+  justify-content: center;
+  margin: 20px auto 0;
+}
+
+input.search {
```
vsm2k17, line 24: all inputs have a class search you can remove `input` from this line
```suggestion
.search {
```
in general, that's a bad idea: don't use tags, use classes for this

vsm2k17 (approved):
good job

PR #514, opened 2021-03-28
Solution

vadym-liubych:
- [DEMO LINK](https://vadym-liubych.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://vadym-liubych.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,16 +1,48 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="UTF-8">
+    <meta charset="UTF-8" >
     <meta
       name="viewport"
       content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
     >
-    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" >
     <title>Search bar airbnb</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./style.css" >
   </head>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -1,16 +1,48 @@
...
+      <label for="big-search" class="form__section">
+        <input
+          type="search"
+          name="big-search"
+          data-qa="keypress"
+          id="big-search"
+          class="form__search big__search"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+        <button class="form__btn big__btn"></button>
```
varseniuk, line 28: Let's be semantic. It should be not a button but just an image

varseniuk (changes requested):
Good work, but let the image to be image (not the button)

src/index.html
vadym-liubych: ok, I'll fix it

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+  background-size: cover;
+  display: block;
+  background-image: url(./images/Search.svg);
+  top: 15px;
+  left: 13px;
+  width: 11px;
+  height: 11px;
+  cursor: pointer;
+  border: none;
+}
```
varseniuk, line 96:
```
.form__section::before,
.form__section-small::before {
  position: absolute;
  background-color: transparent;
  background-repeat: no-repeat;
  background-size: cover;
  display: block;
  cursor: pointer;
  border: none;
}
```

Don't repeat yourself. These are common styles for both `.after`

varseniuk (approved):
Good! 👍 
But don't repeat yourself (read my comment)

PR #662, opened 2021-07-05
Develop

ihor-jpeg:
- [DEMO LINK](https://ihor-jpeg.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ihor-jpeg.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,76 @@
+@font-face {
+  font-family: "Anevir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-weight: normal;
+  font-style: normal;
+}
+
+@font-face {
+  font-family: "Anevir";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-weight: bold;
+  font-style: normal;
+}
+
+body {
+  margin: 0 8px;
+}
+
+.search-container {
+  margin-top: 20px;
+}
+
+.input-bar {
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
+  font-family: "Anevir", Arial, sans-serif;
+  color: #3d4e61;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  color: #3d4e61;
+  font-weight: bold;
+  box-shadow: none;
+}
+
+.input-bar:hover {
+  box-shadow: 0 3px 8px 0 #3d4e6133;
+}
+
+::placeholder {
```
herkoss, line 77: Specify which placeholder(s) should have this style. In your case, all placeholders will have the specified style

herkoss (changes requested):
Well done!👍

But there are some mistakes. Fix it)

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+.small-bar:focus {
+  color: #3d4e61;
+  font-weight: bold;
+
+  background: url(./images/Search.svg) no-repeat,
+    linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  background-position: 13px 49%;
+  background-size: 11px;
+  box-shadow: none;
+}
```
herkoss, line 78: Don't repeat the same styles. This is bad practice. Move them to the common class input-bar

herkoss (changes requested):
Great!👍😊 

But there are one little mistake

src/style.css
ihor-jpeg: Yeah, no clue how I missed that. The fix is here.

herkoss (approved):
Absolutely brilliant💎✨
Keep it up💪

herkoss (approved):

PR #666, opened 2021-07-06
Forms - Homework 2 - Search bar

george-w-boo:
домашка пуста, просто перевіряю, чит npm deploy працює

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@mate-academy/bemlint": "^0.1.1",
     "@mate-academy/eslint-config": "*",
     "@mate-academy/linthtml-config": "0.0.1",
-    "@mate-academy/scripts": "^0.9.1",
+    "@mate-academy/scripts": "^0.9.2",
     "@mate-academy/stylelint-config": "*",
     "backstopjs": "^5.0.1",
     "eslint": "^5.16.0",
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Search bar for Airbnb
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_search-bar-airbnb/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_search-bar-airbnb/report/html_report/)
+- [DEMO LINK](https://george-w-boo.github.io/layout_search-bar-airbnb/)
+- [TEST REPORT LINK](https://george-w-boo.github.io/layout_search-bar-airbnb/report/html_report/)
 
 > Follow [this instructions](https://github.com/mate-academy/layout_task-guideline#how-to-solve-the-layout-tasks-on-github)
 ___
--- a/src/index.html
+++ b/src/index.html
@@ -12,5 +12,6 @@
   </head>
   <body>
     <h1>Search bar airbnb</h1>
+    <p>just checking if npm run deploy function correctly</p>
   </body>
 </html>


src/style.css
```diff
@@ -0,0 +1,87 @@
...
+.search__bar-big {
+  padding: 25px 62px 23px;
+
+  font-size: 16px;
+  line-height: 22px;
+
+  background: url(images/Search.svg) no-repeat 26px 25px;
+}
+
+.search__bar-big.search__bar-big:focus {
```
herkoss, line 52:
```suggestion
.search__bar-big:focus {
```

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  font-weight: bold;
+  box-shadow: none;
+
+  background: url(images/Search.svg) no-repeat 26px 25px,
+    linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
```
herkoss, line 58: Don't repeat the same styles twice. Move them on common class

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+
+.search__bar-small {
+  padding: 12px 33px 10px;
+
+  width: 50%;
```
herkoss, line 63: Search bar must have width 100%, not 50%

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+
+  font-size: 14px;
+  line-height: 19px;
+  
+  text-shadow: 0 3px 2px rgba(0, 0, 0, 0.25);
+  background: url(images/Search.svg) no-repeat 13px 15px;
+  background-size: 11px 11px;
+}
+
+.search__bar-small.search__bar-small:hover {
```
herkoss, line 73:
```suggestion
.search__bar-small:hover {
```

herkoss (changes requested):
Good job!👏👍 

But there are some common mistakes)

src/style.css
george-w-boo: @herkoss this is an exceptional case. If I put `url(images/Search.svg) no-repeat 25px 25px;` to a common class, then this picture will disappear upon focus, because the linear-gradient will cover it ((

src/style.css
george-w-boo: I agree, tnx  

I've just corrected in commit 'version 2.0'

src/style.css
george-w-boo: I agree, tnx  

I've just corrected in commit 'version 2.0'

src/style.css
george-w-boo: I agree, tnx  

I've just corrected in commit 'version 2.0'

src/style.css
herkoss: @george-w-boo You can use properties `background-image: url(u);` and `background-repeat: no-repeat;` in common class but different `background-position: ;`

src/style.css
```diff
@@ -0,0 +1,89 @@
...
+  font-weight: bold;
+  text-shadow: none;
+
+  background: url(images/Search.svg) no-repeat 13px 15px,
+    linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  background-size: 11px 11px, 100% 100%;
+
+  box-shadow: none;
+  outline: none;
+}
```
herkoss, line 89: `search__bar-small:focus` and `search__bar-big:focus` has the same styles. Move it to the new common class. And use this class

herkoss (changes requested):
Well done!)
But there are some common mistakes)

src/style.css
george-w-boo: I tried my best at cutting on repetitive properties. I got rid of a lot of them (look at my commit 'version 3.0'. I also tried dealing with background, and I partly succeeded. Anyway, when I do the thing the used ask here, the needed behavior crashes.

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+}
+
+.sb__form {
+  margin: 20px 0;
+}
+
+.sb__form__small {
+  width: 50%;
```
herkoss, line 28: Width must be 100%, not 50%

herkoss (approved):
Well done!)

PR #690, opened 2021-07-06
search bars created

nastyyvolk:
- [DEMO LINK](https://nastyyvolk.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://nastyyvolk.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,87 @@
+/* fonts */
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: bold;
+  src: url(./fonts/Avenir-Heavy.ttf);
+}
+
+html {
+  box-sizing: border-box;
+  font-family: Avenir, Arial, Helvetica, sans-serif;
+  font-weight: 300;
+  color: #3d4e61;
+}
+
+/* big */
+.search-big {
+  display: block;
+  width: 100%;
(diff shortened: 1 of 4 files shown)


varseniuk (changes requested):
Re-deploy your task to github pages again, please. We've fixed an issue now and deploy should work properly, because now your page looks like this
Снимок экрана 2021-07-06 в 22 11 01

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.search-big::placeholder {
+  font-weight: normal;
+  color: #3d4e61;
+}
+
+/* small */
+.search-small {
+  height: 42px;
```
varseniuk, line 62: You are repeating too many styles in your `search-big` and `search-small` classes. You can create one common class `search` and give at all common classes for both inputs and use it then like `class="search search-big" e.g

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+
+html {
+  box-sizing: border-box;
+  font-family: Avenir, Arial, Helvetica, sans-serif;
+  font-weight: 300;
+  color: #3d4e61;
+}
+
+input.search-big,
+input.search-small {
```
varseniuk, line 22:
```suggestion
.search-big,
.search-small {
```

varseniuk (changes requested):
Don't repeat yourself. Let's create one common class with common styles
Also, pay attention to the selectors

varseniuk (approved):
Good job!

PR #761, opened 2021-08-16
search-bar solution

s-person27:
- [DEMO LINK](https://s-person27.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://s-person27.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,108 @@
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  src: url(.\fonts\Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 700;
+  src: url(.\fonts\Avenir-Heavy.ttf);
+}
+
+html,
+*::before,
+*::after {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+  font-size: 16px;
+  line-height: 22px;
+}
+
+.search-form__big {
+  margin: 20px 0;
+}
+
+.search-bar {
+  width: 100%;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,108 @@
...
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 700;
+  src: url(.\fonts\Avenir-Heavy.ttf);
+}
+
+html,
+*::before,
+*::after {
```
herkoss, line 15:
```suggestion {
```

herkoss: Don't write like this. It is a bad practice

herkoss: You can write like this: 
\* {
  box-sizing: border-box;
}

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  line-height: 19px;
+  color: #3d4e61;
+}
+
+.search-form__small-input::placeholder {
+  font-family: Avenir, sans-serif;
+  font-size: 14px;
+  line-height: 19px;
+  color: #3d4e61;
+}
```
herkoss, line 108: Don't repeat yourself. If the same styles are used on different elements, then you need to bring these styles into a common class and use this class in both elements.

herkoss (changes requested):
Well done!👍
But it needs some refactoring)
Good luck🍀

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+.search-form__small-input {
+  height: 42px;
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 19px;
+  background-image: url("./images/Search.svg");
+  background-repeat: no-repeat;
+  background-size: 11px;
+  background-position: 12px 14px;
+}
```
herkoss, line 82: A property `background-image: url("./images/Search.svg");` is duplicated. Avoid it

herkoss (changes requested):
Almost done!)

herkoss (approved):
Good work!)

PR #763, opened 2021-08-16
Search-Bar Solution

Daerniss:


src/style.css
```diff
@@ -0,0 +1,114 @@
+* {
+  box-sizing: inherit;
+  padding: 0;
```
Ihor-Karpyn, line 3:
```suggestion
```

Don't reset padding for all.  It's very difficult operation

Ihor-Karpyn:
```
p, h1, ... {
  padding: 0;
  margin: 0;
}
```

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  font-family: "Avenir";
+  font-weight: bold;
+  font-style: normal;
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
+body {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+  box-sizing: border-box;
```
Ihor-Karpyn, line 23:
```suggestion
```

Ihor-Karpyn (changes requested):
Good job 👌, let's fix few comments

src/style.css
```diff
@@ -0,0 +1,108 @@
...
```
Ihor-Karpyn, line 22:
```suggestion
```

again

Ihor-Karpyn (approved):
Great, pay attention to prevue comments

src/style.css
Daerniss: But how do I keep border-box for every element then?

PR #805, opened 2021-08-22
readme.md

cTATIa:
[DEMO LINK](https://cTATIa.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://cTATIa.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,76 @@
+@font-face {
+  font-family: "Avenir";
+  src: url("./fonts/Avenir.ttc");
+}
+
+@font-face {
+  font-family: "Avenir-Book";
+  src: url("./fonts/Avenir-Book.ttf");
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
+html {
+  font-weight: 300;
+  font-size: 16px;
+}
+
+body {
+  margin: 0;
+  padding: 20px 8px;
+}
+
+.form {
+  position: relative;
+  margin-bottom: 20px;
+}
+
+.search {
+  font-family: Avenir, sans-serif;
+  width: 100%;
+  color: #3d4e61;
+  background: #fff;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  padding: 20px 8px;
+}
+
+.form {
+  position: relative;
+  margin-bottom: 20px;
+}
+
+.search {
+  font-family: Avenir, sans-serif;
```
aliceblunt8, line 32: add font-family to html

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  width: 100%;
+  color: #3d4e61;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
```
aliceblunt8, line 37: we usually add box-sizing to * elements

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  background-repeat: no-repeat;
+  background-position: 26px 25px;
+  background-size: 19px 19px;
+  padding-left: 62px;
+}
+
+.small-search {
+  height: 42px;
+  font-style: normal;
+  font-weight: normal;
```
aliceblunt8, line 56: let's add these styles to html too

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
+html {
+  font-weight: 300;
+  font-size: 16px;
```
aliceblunt8, line 18: no need to add font-size to html, cause then each element has it's own font-size
add to html only common styles

aliceblunt8 (changes requested):

aliceblunt8 (approved):
you've just deleted box-sizing: border-box, but you need to add it with * selector (so it can be applied for ALL elements)

PR #845, opened 2021-09-14
[Checklist]: Expand checklist and readme

VinnieJ-2k20:
Update readme - add hints for frequent mistakes.
Update checklist - add code style examples and hints for tests.

### DIFF
--- a/checklist.md
+++ b/checklist.md
@@ -185,4 +185,6 @@ BAD example:
 input could be activated by clicking on the corresponding label.
 8. [TESTS] - Remember to style input placeholders correctly
 9. [TESTS] - Remember that the search-bar should have width: 100% on any screen.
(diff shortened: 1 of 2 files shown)


checklist.md
```diff
@@ -1,6 +1,188 @@
...
+  id="dateOfBirth"
+  required
+>
+```
+
+7. [FUNCTIONALITY] - You need to use a label tag for each input, so that every
+input could be activated by clicking on the corresponding label.
+8. [TESTS] - Remember to style input placeholders correctly
+9. [TESTS] - Remember that the search-bar should have width: 100% on any screen.
+
```
solaryasha, line 188: Maybe, we should add rule that each input should have separate form

solaryasha (approved):

PR #856, opened 2021-09-25
write code

Batick3737:


src/index.html
```diff
@@ -11,6 +11,37 @@
...
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action=""
+      data-qa="big"
+      method="GET"
+      class="margin"
+    >
+      <label class="position1" for="1">
```
qoa11a, line 20: Give more clear class names. For example small-input, big-input, input, and label.
Don't use just numbers as an ID. Give proper names.

qoa11a (changes requested):

src/style.css
```diff
@@ -0,0 +1,110 @@
...
+
+.big-input:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
+  border-radius: 3px;
+  font-weight: 900;
+}
+
+.small-input {
```
polosanya, line 56: You have similar properties for both small and big inputs. Give them a common class (e.g. "input") and move them under its selector
image

polosanya (changes requested):
Almost done!
Don't forget to add demo and test links to your PR description, so mentors can faster check your work
image
1. Placeholder shoudn't change style on focus
image
2. Font style on focus doesn't match the design
image
image

Batick3737:
https://batick3737.github.io/layout_search-bar-airbnb/

src/index.html
```diff
@@ -11,6 +11,37 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action=""
+      data-qa="big"
+      method="GET"
+      class="margin"
```
agniya-i, line 18:
```suggestion
      class="form"
```
it would be better to not name classes by properties in this case, what if you need to add other properties to this class?)

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+.label-small::before {
+  top: 3px;
+  left: 13px;
+  height: 11px;
+  width: 11px;
+}
+
+.input {
+  font-family: "Avenir", Helvetica, sans-serif;
+  font-style: normal;
```
agniya-i, line 31: `normal` is a default value for `font-style`, so you can remove this

src/style.css
```diff
@@ -0,0 +1,70 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir.ttc);
+}
```
agniya-i, line 4: add `font-weight` and `font-style`
Screenshot 2023-04-18 at 17 44 51

agniya-i (changes requested):

PR #1021, opened 2021-12-17
Develop Serhii-Khobotov

Serhii-Khobotov:


danylo-onishchuk (changes requested):
при фокусе не должно быть аутлайна, а также должен меняться беграунд

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+}
+
+.searchSmall:hover {
+  background-image: url("images/Search.svg");
+  height: 42px;
+  padding-left: 33px;
+  background-repeat: no-repeat;
+  background-position: 13px center;
+  background-size: 11px;
+}
```
aliceblunt8, line 90: these styles shouldn't copy styles from searchSmall and searchBig classes
add box-shadow on hover and then change background and text styles on focus

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+
+.searchBig:focus {
+  height: 70px;
+  background-image: url("images/Search.svg");
+  padding-left: 62px;
+  background-repeat: no-repeat;
+  background-position: 26px center;
+  background-size: 19px;
+}
```
aliceblunt8, line 99: check with design, you need to add `background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);` on focus and other styles to make your default input look like this one when focused
Screenshot 2021-12-29 at 19 04 50

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.search:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
+  border-radius: 3px;
+}
```
aliceblunt8, line 72: let's leave only this styles for hover and focus, cause the ones you wrote after them can overlap these styles and they don't apply at all

aliceblunt8 (dismissed):

varseniuk (approved):

PR #1045, opened 2021-12-21
created search-bar

Dorosh13:
https://Dorosh13.github.io/layout_search-bar-airbnb/
https://Dorosh13.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,26 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form method="post" data-qa="big">
+      <div class="big-form">
+        <label for="big" class="img-label1"></label>
+        <input
+          type="text"
+          id="big"
+          placeholder="Try &#34Los Angeles&#34"
+          data-qa="keypress"
+        >
+      </div>
+    </form>
+    <form method="post" data-qa="small">
+      <div class="small-form">
+        <label for="small" class="img-label2"></label>
+        <input
+          type="text"
+          id="small"
+          placeholder="Try &#34Los Angeles&#34"
+        >
+      </div>
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,122 @@
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  font-style: normal;
+  font-weight: 900;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+  color: #404e5f;
+}
+
+#big {
+  font-family: "Avenir", sans-serif;
+  width: 100%;
+  height: 100%;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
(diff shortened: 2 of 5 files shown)


src/index.html
```diff
@@ -11,6 +11,26 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form method="post" data-qa="big">
```
Anna-Tkachenko, line 14: You should add action attribute to the form

src/index.html
```diff
@@ -11,6 +11,26 @@
...
+        <label for="big" class="img-label1"></label>
+        <input
+          type="text"
+          id="big"
+          placeholder="Try &#34Los Angeles&#34"
+          data-qa="keypress"
+        >
+      </div>
+    </form>
+    <form method="post" data-qa="small">
```
Anna-Tkachenko, line 25: Same here

src/style.css
```diff
@@ -0,0 +1,122 @@
...
+  font-weight: 900;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+  color: #404e5f;
+}
+
+#big {
```
Anna-Tkachenko, line 19: Use class names for setting styles

src/style.css
```diff
@@ -0,0 +1,122 @@
...
+  box-sizing: border-box;
+  border-radius: 3px;
+}
+
+::placeholder {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+  line-height: 22px;
+  color: #3d4e61;
+}
```
Anna-Tkachenko, line 52: Set correct font-weight property to inputs and placeholders

Anna-Tkachenko (changes requested):
Use class names for setting styles and add correct font-weight for inputs and placeholders (you added this property only in font-face but didn't set to elements)

Anna-Tkachenko (changes requested):
Search bar should have 100% width

Anna-Tkachenko (changes requested):
Did you run npm run deploy after changes? Search bar in demo-link still doesn't have 100% width

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+
+html {
+  font-family: "Avenir", sans-serif;
+  color: #404e5f;
+}
+
+.big {
+  font-family: "Avenir", sans-serif;
+  width: 100%;
+  height: 100%;
```
Anna-Tkachenko, line 22: You can also use font-family: inherit for inputs

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  width: 19px;
+  height: 19px;
+  margin: 25px 0 26px 26px;
+  background-image: url(images/Search.svg);
+  background-size: 100%;
+}
+
+.small {
+  font-family: "Avenir", sans-serif;
+  background: #fff;
```
Anna-Tkachenko, line 72: Same here

Anna-Tkachenko (changes requested):
Add one main class for all inputs and set general css properties to it. Don't repeat all styles for 2 different classes

Anna-Tkachenko (approved):
You didn't add one general class for both of forms. Add it, when you'll have time
Classes .big and .small should contain styles that can be different for forms, new class (.form, for example) should contain styles that are the same for small and big forms

PR #1052, opened 2021-12-21
'create-bars'

mdanko990:
- [DEMO LINK](https://mdanko990.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://mdanko990.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,64 @@
+@font-face {
+  font-family: "Avenir-book";
+  src: url("./fonts/Avenir-Book.ttf");
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir-heavy";
+  src: url("./fonts/Avenir-Heavy.ttf");
+  font-weight: bold;
+}
+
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  width: 100%;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  outline: none;
+  font-family: "Avenir-book", sans-serif;
+}
```
Anna-Tkachenko, line 33: Will be better add default font to html tag and use needed font-weight for inputs and for ::placeholder

Anna-Tkachenko (approved):
Good job! 👍 
Will be better to set correct font-weight for all inputs and set correct font styles to placeholder (using ::placeholder)

PR #1089, opened 2022-01-18
Add index.html and style.css

GoDfreE25:


src/style.css
```diff
@@ -0,0 +1,69 @@
...
+
+.bigBar::placeholder {
+  color: rgba(61, 78, 97, 1);
+  font-size: 16px;
+  font-weight: 400;
+}
+
+.smallBar::placeholder {
+  color: rgba(61, 78, 97, 1);
+  font-size: 14px;
```
solaryasha, line 55: Do you need this rule ?

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+  background: url(./images/Search.svg) no-repeat;
+  background-position: 13px;
+  background-size: 11px;
+  font-size: 14px;
+  padding-left: 33px;
+}
+
+.bigBar::placeholder {
+  color: rgba(61, 78, 97, 1);
+  font-size: 16px;
```
solaryasha, line 49: Do you need this rule ?

solaryasha (changes requested):
When i focus small search bar - icon and text are come together (you should fix it)
Screenshot 2022-01-18 at 15 17 06

src/style.css
GoDfreE25: Yes

src/style.css
GoDfreE25: Yes

GoDfreE25: for small i need , for big i don't need it

GoDfreE25: i need because i fail 2 npm tests

GoDfreE25: need for fonts

src/style.css
solaryasha: I guess, you could remove it

solaryasha (approved):

src/style.css
GoDfreE25: I need this for fonts too

PR #1092, opened 2022-01-18
First commit SearchBar

Olexandr-Trubin-Koshyk:
- [DEMO LINK](https://olexandr-trubin-koshyk.github.io//layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://olexandr-trubin-koshyk.github.io//layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,76 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+.labelSearchBar {
+  display: block;
+  position: relative;
+}
+
+.searchBar {
+  font-family: Avenir, sans-serif;
+  font-size: 16px;
```
solaryasha, line 25: It's better to move these  rules to `html` tag and then inherit here

solaryasha (approved):
Good job !

PR #1206, opened 2022-02-16
test3

dovzh1:


src/style.css
```diff
@@ -0,0 +1,102 @@
+* {
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: Avenir-Book;
+  src: url(fonts/Avenir-Book.ttf);
```
illiarerun, line 7: you should connect 3 fonts for 3 different font-weights

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+}
+
+body {
+  padding-top: 12px;
+}
```
illiarerun, line 12: ?

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+
+.field {
+  font-family: Avenir-Book, sans-serif;
+  font-weight: 900;
```
illiarerun, line 16: should be 300 by default and 900 on focus

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+  height: 100%;
+  width: 100%;
+  border: none;
+  outline: none;
+}
+
+.fieldBig {
```
illiarerun, line 23: use snake-case, not the camelCase for class names

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+}
+
+.fieldSmall {
+  font-weight: 400;
+  font-size: 14px;
+  padding-left: 33px;
+}
+
+.field::placeholder {
+  font-family: Avenir-Book, sans-serif;
```
illiarerun, line 35: redundant
```suggestion
```

illiarerun (changes requested):
Nice start, now let's work on fixes:

src/style.css
dovzh1: done

src/style.css
dovzh1: I had to apply it. Otherwise, there was a small distance to the top and the test was not passed...

src/style.css
dovzh1: done

src/style.css
dovzh1: done

src/style.css
dovzh1: done

MaksimusCRS (approved):
Well done!

PR #1225, opened 2022-03-14
created html and css and modified readme

AndreiSuc:
- [DEMO LINK](https://andreisuc.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://andreisuc.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,99 @@
+@font-face {
+  font-family: "Avenir-Book";
+  src: url("./fonts/Avenir-Book.ttf");
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
+.page {
+  --big-bar-height: 70px;
+  --big-bar-pleft: 62px;
+  --big-bar-border: 1px;
+  --big-search-icon-top: calc(var(--big-bar-height) / 2);
+
+  --small-bar-height: 42px;
+  --small-bar-pleft: 33px;
+  --small-bar-border: 1px;
+  --small-search-icon-top: calc(var(--small-bar-height) / 2);
+
+  font-family: "Avenir-Book", sans-serif;
+  font-weight: 300;
+}
+/*
+.page-body {
+  margin: 0;
+  padding: 0;
+} */
+
+.container {
+  margin-top: 20px;
+  position: relative;
+  display: block;
+  height: var(--big-bar-height);
+}
+
+.container--small {
+  height: var(--small-bar-height);
+}
+
+.search-icon {
+  position: absolute;
+  top: var(--big-search-icon-top);
+  left: 26px;
+  transform: translateY(-50%);
+  width: 19px;
+  height: 19px;
+}
+
+.search-icon--small {
+  top: var(--small-search-icon-top);
+  left: 13px;
+  width: 11px;
+  height: 11px;
+}
+
+.bar {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background: #fff;
+  border: var(--big-bar-border) solid #e1e7ed;
+  box-sizing: border-box;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  font-size: 16px;
+  width: 100%;
+  font-family: "Avenir-Book", sans-serif;
+  font-weight: 300;
+
+  display: block;
+  padding-left: var(--big-bar-pleft);
+  height: var(--big-bar-height);
+}
+
+.bar--small {
+  font-size: 14px;
+  padding-left: var(--small-bar-pleft);
+  height: var(--small-bar-height);
+}
+
+.bar:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.bar:focus {
+  box-shadow: 0 0 0 rgba(61, 78, 97, 0.2);
+  font-family: "Avenir-Heavy", sans-serif;
+  border-radius: 3px;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
+
+.bar::placeholder {
+  color: #3d4e61;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,99 @@
...
+  height: 11px;
+}
+
+.bar {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background: #fff;
```
ihor-jpeg, line 64: you do not really need position absolute here. it will naturally be in its place
bg-color is white by default as well

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+  border: var(--big-bar-border) solid #e1e7ed;
+  box-sizing: border-box;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  font-size: 16px;
+  width: 100%;
+  font-family: "Avenir-Book", sans-serif;
+  font-weight: 300;
+
+  display: block;
```
ihor-jpeg, line 74: i dont think you need display block here. it does not do anything

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+  padding-left: var(--big-bar-pleft);
+  height: var(--big-bar-height);
```
ihor-jpeg, line 76: instead of setting the height explicitly, try to set the correct padding, font size, and line-height. the height will then be calculated automatically

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+}
+
+.bar:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.bar:focus {
+  box-shadow: 0 0 0 rgba(61, 78, 97, 0.2);
+  font-family: "Avenir-Heavy", sans-serif;
```
ihor-jpeg, line 92: use one font-family for the entire page. you only need to set it in one place

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+
+  display: block;
+  padding-left: var(--big-bar-pleft);
+  height: var(--big-bar-height);
+}
+
+.bar--small {
+  font-size: 14px;
+  padding-left: var(--small-bar-pleft);
+  height: var(--small-bar-height);
```
ihor-jpeg, line 82: avoid setting dimensions exlicitly

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+
+.page {
+  --big-bar-height: 70px;
+  --big-bar-pleft: 62px;
+  --big-bar-border: 1px;
+  --big-search-icon-top: calc(var(--big-bar-height) / 2);
+
+  --small-bar-height: 42px;
+  --small-bar-pleft: 33px;
+  --small-bar-border: 1px;
```
ihor-jpeg, line 19: this is the same as `--big-bar-border`. no need to repeat it

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+
+.container {
+  margin-top: 20px;
+  position: relative;
+  display: block;
+  height: var(--big-bar-height);
+}
+
+.container--small {
+  height: var(--small-bar-height);
```
ihor-jpeg, line 39: i dont think you need to hardcode the height. i explained in one of the comments

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+/*
+.page-body {
+  margin: 0;
+  padding: 0;
+} */
+
+.container {
+  margin-top: 20px;
+  position: relative;
+  display: block;
```
ihor-jpeg, line 34: your container is a `div`. its display prop. is set to block by default

ihor-jpeg (changes requested):
Good work. Let's fix a few problems. You seem to be writing quite a bit of unnecessary code. Try to clean it up and only use it if you have to. Also, do not position everything using absolute. It's okay for the icon, though. Another option for the icon would be setting it as a background image

src/style.css
AndreiSuc: I used this approach in the beginning. But the bg-color should become linear-gradient on focus, which overlays the background image of the icon.

AndreiSuc: > you do not really need position absolute here. it will naturally be in its place bg-color is white by default as well

I used this approach in the beginning. But the bg-color should become linear-gradient on focus, which overlays the background image of the icon.

src/style.css
AndreiSuc: Okay, removed.

src/style.css
AndreiSuc: I have to set the height because the clearance from the top for my icons depends on it. When I set the height by padding and font size I to hardcode the position of my absolutely placed icons.

src/style.css
AndreiSuc: Ok, changed.

src/style.css
AndreiSuc: Fixed.

src/style.css
AndreiSuc: Fixed.

src/style.css
AndreiSuc: I tried to get rid of the height, but then I have no anchor for my absolutely placed icons.

AndreiSuc:
I have to use height because I have images that are absolutely placed and their clearance from the top is based on the height of the search bars. And I can't set them as a background of the input because when the input becomes focused its background changes to linear-gradient, which substitutes the icon set as a background.

src/style.css
ihor-jpeg: I'm not sure what you mean.
I have disabled the positioning and the bg-color and it seems to work fine. 
Screenshot 2022-03-15 at 10 42 48

Screenshot 2022-03-15 at 10 42 13


One thing I have noticed is that when I click on the big search bar, the shadow on the small one is altered

src/style.css
ihor-jpeg: if the icons are positioned using absolute, don't they already have hardcoded values for top and left?

AndreiSuc: Okay, I fixed.

ihor-jpeg (changes requested):
Your code looks good! Let's fix this little thingy.
Screenshot 2022-03-18 at 12 52 54

ihor-jpeg (approved):
Noooice! 👍

PR #1226, opened 2022-03-15
the layout of searchbar is created, css codes are iplemented, 

efdalyalcin:
there are problems with small searchbar, I am tired to solve them so I will work on them later at night

- [DEMO LINK](https://efdalyalcin.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://efdalyalcin.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,36 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="#"
+      method="get"
+      class="bar"
+      data-qa="big"
+    >
+      <label for="search-1"></label>
+      <input
+        type="search"
+        name="search-1"
+        id="search-1"
+        class="search"
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+      >
+    </form>
+    <form
+      action="#"
+      method="get"
+      class="bar"
+      data-qa="small"
+    >
+      <label for="search-2"></label>
+      <input
+        type="search"
+        name="search-2"
+        id="search-2"
+        class="search small-class"
+        placeholder="Try &quot;Los Angeles&quot;"
+      >
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,47 @@
+html {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.bar {
+  margin: 20px 8px;
+}
+
+.search {
+  width: 100%;
+  height: 70px;
+  padding: 5px 62px;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-position: 25px 50%;
+  background-size: 20px;
+  font-size: 16px;
+  line-height: 22px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  font-family: "Avenir-Book", sans-serif;
+  font-weight: 300;
+}
+
+.small-class {
+  font-size: 14px;
+  height: 42px;
+  background-position: 25px 50%;
+  background-size: 14px;
+}
+
+.search:hover {
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+}
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Search bar for Airbnb
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_search-bar-airbnb/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_search-bar-airbnb/report/html_report/)
+- [DEMO LINK](https://efdalyalcin.github.io/layout_search-bar-airbnb/)
+- [TEST REPORT LINK](https://efdalyalcin.github.io/layout_search-bar-airbnb/report/html_report/)
 
 > Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/#how-to-solve-the-layout-tasks-on-github)
 


VinnieJ-2k20 (changes requested):
You are on the right way.
But still need to fix couple of things.
1. Magnifying glass disappears on hover and focus.
2. Font of the entered text is different from the mockup.
image

src/style.css
```diff
@@ -0,0 +1,59 @@
...
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-position: 25px 50%;
+  background-size: 20px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  line-height: 22px;
+  font-size: 16px;
+  font-family: "Avenir-Book", sans-serif;
```
meforoma, line 26: you haven't imported the font, so apparently proper font-family is not applied - and font is using a fallback sans-serif atm
```
@font-face {
  font-family: "Avenir";
  ...
}
```
okok?

src/style.css
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.search:hover {
+  background-color: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  background-color: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
meforoma, line 45: `linear-gradient` wouldn't work with `background-color`.
You should use `background` shorthand, but consider specifics of it's syntax.
Basically you'd have to put both gradient and icon image inside and state alignment and object-fit.
Read about it here: https://developer.mozilla.org/en-US/docs/Web/CSS/background
this should give you the idea:
Snapshot

you'll need to also position and size your background (`background-position / -size`)

src/style.css
```diff
@@ -0,0 +1,59 @@
...
+  font-size: 16px;
+  font-family: "Avenir-Book", sans-serif;
+  font-weight: 300;
+}
+
+.small-class {
+  font-size: 14px;
+  height: 42px;
+  padding: 5px 33px;
+  background-position: 12px 48%;
```
meforoma, line 34: hmm, why 48% in particular? (here and later in the code)
tried to pixel-match with a reference?

meforoma (changes requested):

src/style.css
efdalyalcin: Yes, i pixel-match with the reference. I didn't change it in the new push, should I?

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  line-height: 22px;
+  font-size: 16px;
+  font-family: "Avenir-Book", sans-serif;
+  font-weight: 300;
+}
+
+.small-class {
+  font-size: 14px;
+  height: 42px;
+  padding: 5px 33px;
```
meforoma, line 43: `#boxed-paddings`
same issue here - use only `padding-left`

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.search {
+  width: 100%;
+  height: 70px;
+  padding: 5px 62px;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-position: 25px 50%;
+  background-size: 20px;
```
meforoma, line 30: why 20? mock sets 19px

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  margin: 20px 8px;
+}
+
+.search {
+  width: 100%;
+  height: 70px;
+  padding: 5px 62px;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-position: 25px 50%;
```
meforoma, line 29: no actual need for the 50%
- only 25px works just the same
```suggestion
  background-position: 25px;
```

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  font-size: 16px;
+  font-family: "Avenir-Book", sans-serif;
+  font-weight: 300;
+}
+
+.small-class {
+  font-size: 14px;
+  height: 42px;
+  padding: 5px 33px;
+  background-position: 12px 48%;
```
meforoma, line 44: no need for the second measure
```suggestion
  background-position: 12px;
```

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+
+.small-class:hover {
+  background-position: 12px 48%;
+  background-size: 11px;
+}
+
+.small-class:focus {
+  background-position: 12px 48%;
+  background-size: 11px;
+}
```
meforoma, line 70: you're duplicating the code, combine like this:
and no percentages.
```suggestion
.small-class:hover,
.small-class:focus {
  background-position: 12px;
  background-size: 11px;
}
```

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  background: no-repeat url(images/Search.svg)
+    linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+  font-family: "Avenir-Heavy", sans-serif;
+}
```
meforoma, line 60: a set of comments here
1. background shorthand syntax is wrong - you've missing a comma to split image and gradient
2. border and border-radius duplicate the `.search` style
3. you don't need an outline on focus
4. position and size background(`-size/-position`)

```suggestion
.search:focus {
  background: center / contain no-repeat url(images/Search.svg),
    linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
  /* border: 1px solid #e1e7ed;
  border-radius: 3px; both are duplicating the already set style for .search */
  font-family: "Avenir-Heavy", sans-serif;
  outline: none;
  background-size: 19px;
  background-position: 25px;
}
```

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  padding: 5px 33px;
+  background-position: 12px 48%;
+  background-size: 11px;
+}
+
+.search:hover {
+  background-color: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
```
meforoma, line 52: color and border change nothing - redundant code
```suggestion
.search:hover {
  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
}
```

meforoma (changes requested):
please apply changes , commit-push and `npm run deploy`

src/index.html
```diff
@@ -11,6 +11,36 @@
...
+      method="get"
+      class="bar"
+      data-qa="small"
+    >
+      <label for="search-2"></label>
+      <input
+        type="search"
+        name="search-2"
+        id="search-2"
+        class="search small-class"
```
ihor-jpeg, line 41: it's better to give this class the name of `search--small`, but you also need to have `search--big`.
at the moment you are styling everything under the `search` class and this does not make a lot of sense. for a better understanding of what is happening, I suggest you have a class `search` and both search bars have this class. it contains the common styles. and you also have `search--small` and `search--big` which contain classes that differ. i think it will be a lot more structured

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  outline: none;
+  background-size: 19px;
+  background-position: 25px;
+}
+
+.small-class:hover,
+.small-class:focus {
+  background-size: 11px;
+  background-position: 12px;
+}
```
ihor-jpeg, line 66: this has the same values as the `small-class` why do you need to specify it here again?

ihor-jpeg (changes requested):
Almost there!

PR #1227, opened 2022-03-15
added 2 searchbars, WIP

LukaszGryziewicz:


src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  font-weight: 300;
+}
+
+.searchbar {
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  width: 100%;
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
```
meforoma, line 33: @LukaszGryziewicz search icons disappears when typing text inside input box.
try applying `background-image` to an `::after` pseudo-element:
```
.searchbar::after {
  ...
}
```

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  margin-top: 20px;
+}
+
+::placeholder {
+  font-family: "Avenir-Book", sans-serif;
+  font-size: 16px;
+  font-weight: 300;
+}
+
+.searchbar {
```
meforoma, line 27: if you pay closer attention to the text color you'll find it's not a default one.
Please take a closer look.

meforoma (changes requested):

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  background-repeat: no-repeat;
+}
+
+.searchbar:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.searchbar:focus {
+  border-radius: 3px;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
meforoma, line 44: for proper layout and effects you need to use combined shorthand syntax for both image and image:
Also you don't need an outline on focus:
```suggestion
  background: center / contain no-repeat url(./images/Search.svg),
    linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
  outline: none;
```

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  background-size: 19px 19px;
+  background-position: 26px center;
+}
+
+.big-searchbar:focus {
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: 19px 19px;
+  background-position: 26px center;
+}
```
meforoma, line 61: when using an earlier mentioned combined shorthand syntax for image and color, you can ease up a bit:
`background-size: 19px 19px;` second 19px will cause gradient to multi-line, so remove it.
```suggestion
.big-searchbar:focus {
  background-size: 19px;
  background-position: 26px center;
}
```

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  background-position-x: 12px;
+}
+
+.small-searchbar:focus {
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: 11px 11px;
+  background-position-y: 14px;
+  background-position-x: 12px;
+}
```
meforoma, line 77: similar to `.big-searchbar:focus {`
```suggestion
.small-searchbar:focus {
  background-size: 11px 11px;
  background-position: 15px center;
}
```

meforoma (changes requested):
please apply all fixes and re-deploy your solution and tests.

ihor-jpeg (approved):
Approve, so you can move forward on the platform, but please fix the comment.

PR #1228, opened 2022-03-15
completed search bar task

59MrRobot:
- [DEMO LINK](https://59MrRobot.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://59MrRobot.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" class="page">
```
VinnieJ-2k20, line 2: You don't use this class anywhere. Therefore, it's useless.
```suggestion
<html lang="en">
```

src/index.html
```diff
@@ -10,7 +10,40 @@
...
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <div class="container">
+      <form
+        action="/"
+        method="get"
+        data-qa="big"
+      >
+        <label for="big-search">
```
VinnieJ-2k20, line 21: If you add `for` attribute (which is good), you need to add corresponding `id` to the input.

VinnieJ-2k20 (changes requested):
You are on the right way.
But couple of things need fixing:
1. Text color is not plain black. Check the mockup for more info.
2. You are missing gradient background on focus. Check corresponding pseudo-class selectors and what styles should be overriden
image

src/index.html
59MrRobot: I've removed it.

src/index.html
59MrRobot: Noted, thank you.

meforoma (approved):

meforoma (approved):
nicely done 👍

PR #1243, opened 2022-03-17
This is where I am. WIP

proznerka:
- [DEMO LINK](https://proznerka.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://proznerka.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,62 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 20px;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir-Book, Avenir-Heavy;
+  src: url(fonts/Avenir-Book.ttf);
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+.form {
+  margin-top: 20px;
+  display: inline-block;
+  width: 100%;
+
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+}
+
+.bigInput {
+  display: inline-block;
+  border: none;
+  margin: 10px;
+  margin-left: 40px;
+  width: calc(100% - 150px);
+  height: 70px;
+}
+
+.smallInput {
+  display: inline-block;
+  border: none;
+  margin-left: 30px;
+  width: calc(100% - 150px);
+  height: 42px;
+}
+
+.bigImage {
+  background-position: absolute;
+  display: inline-block;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  width: 19px;
+  height: 19px;
+  margin: 25px;
+}
+
+.smallImage {
+  background-position: absolute;
+  display: inline-block;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  width: 11px;
+  height: 11px;
+  margin: 13px 15px;
+}
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@mate-academy/bemlint": "^0.1.1",
     "@mate-academy/eslint-config": "*",
     "@mate-academy/linthtml-config": "0.0.2",
-    "@mate-academy/scripts": "^0.9.13",
+    "@mate-academy/scripts": "^0.9.14",
     "@mate-academy/stylelint-config": "*",
     "backstopjs": "^5.0.1",
     "eslint": "^5.16.0",
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Search bar for Airbnb
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_search-bar-airbnb/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_search-bar-airbnb/report/html_report/)
+- [DEMO LINK](https://proznerka.github.io/layout_search-bar-airbnb/)
+- [TEST REPORT LINK](https://proznerka.github.io/layout_search-bar-airbnb/report/html_report/)
 
 > Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/#how-to-solve-the-layout-tasks-on-github)
 
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,39 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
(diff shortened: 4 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,111 @@
...
+  font-family: Avenir-Heavy, sans-serif;
+}
+
+.bigInput:focus::placeholder {
+  color: transparent;
+}
+
+.smallInput:focus::placeholder {
+  color: transparent;
+}
```
ihor-jpeg, line 107: placeholder does not inherit anything. so set the font-size and font-family here explicitly

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+
+.form {
+  display: inline-block;
+  width: 100%;
+  margin-bottom: 20px;
+  background: #fff;
+  /* border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px; */
```
ihor-jpeg, line 35: let's remove this, there should be a border only on the input
box-shadow as well, only on the input

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+html {
+  font-family: Avenir-Book, sans-serif;
+  color: rgba(61, 78, 97, 1);
+  font-weight: 300;
+  font-size: 16px;
+}
+
+body {
+  margin: 20px 10px 0;
+}
```
ihor-jpeg, line 26: do not set margin here and do not reset it

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+
+.form {
+  display: inline-block;
```
ihor-jpeg, line 29: do not set display prop here

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+  width: 100%;
+  margin-bottom: 20px;
```
ihor-jpeg, line 31: use margin-top

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+
+.bigInput {
+  display: inline-block;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  width: 100%;
+  /* height: 70px; */
+  padding: 24px 0;
+  line-height: 20px;
```
ihor-jpeg, line 46: using the height here is alright. but remove line-height and padding

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+.bigImage {
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-position-y: center;
+  background-position-x: 26px;
+  text-indent: 62px;
```
ihor-jpeg, line 54: instead of text-indent here use the same padding (62px) for .bigInput

ihor-jpeg (changes requested):
I have left a few comments. See if it helps

src/style.css
proznerka: You changed the comment after I got how to do it 🤣

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+
+.bigInput::placeholder {
+  font-family: Avenir-Book, sans-serif;
+  font-size: 16px;
+}
+
+.bigImage {
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-position-y: center;
```
ihor-jpeg, line 55: you are using the exact same styles for two search bars. let's create a class `searchBar` or anything you like, that will have all the common styles for two searchbars. then create two modifiers of that class, for example `searchBar--big` and `searchBar--small`. these will have everything that is different. you can also remove separate classes for images and move everything under that one `searchBar` class and the modifiers. it will be much easier to work with

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  height: 70px;
+  padding-left: 62px;
+  /* padding: 24px 0;
+  line-height: 20px; */
+}
+
+.bigInput::placeholder {
+  font-family: Avenir-Book, sans-serif;
+  font-size: 16px;
+}
```
ihor-jpeg, line 50: same here, create common styles, like `font-family` for all placeholders, and then put styles that differ for each input's placeholder

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  display: inline-block;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  width: 100%;
+  height: 70px;
+  padding-left: 62px;
+  /* padding: 24px 0;
+  line-height: 20px; */
+}
```
ihor-jpeg, line 45: structure your styles according to what they do, it will be easier to work with

```suggestion
.bigInput {
  display: inline-block;
  width: 100%;
  height: 70px;
  padding-left: 62px;
  
  border: 1px solid #e1e7ed;
  border-radius: 4px;
  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
}
```

ihor-jpeg (changes requested):
Okay. That's already much better. Now let's do some cleanup, and we will figure out what is wrong

src/style.css
ihor-jpeg: sry 😆

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+  outline: none;
+  font-family: Avenir-Heavy, sans-serif;
+}
+
+.searchBar--small:focus {
+  background-image: url(images/Search.svg), linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  background-repeat: no-repeat, repeat;
+  border: 1px solid rgba(225, 231, 237, 1);
+  border-radius: 3px;
+  outline: none;
```
ihor-jpeg, line 83: generally, if you remove the outline, you need to set a special border on focus, so a user has an idea of which search bar is selected. here the background changes, but just remember it for the future

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+  background-size: 11px;
+}
+
+.searchBar--big::placeholder {
+  font-family: Avenir-Book, sans-serif;
+}
+
+.searchBar--small::placeholder {
+  font-family: Avenir-Book, sans-serif;
+}
```
ihor-jpeg, line 63:
```suggestion
.searchBar::placeholder {
  font-family: Avenir-Book, sans-serif;
}
```

ihor-jpeg (approved):
Well done! 🚀  You did it! Yay! But take a look at the comments

src/style.css
ihor-jpeg: also, i think we need to set the font color from figma for the placeholders

PR #1250, opened 2022-03-18
Develop

Sardorbek1122:
- [DEMO LINK](https://Sardorbek1122.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Sardorbek1122.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,42 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <!-- search bar starts here -->
+    <div class="container-bar">
+      <form action="/" method="get" data-qa="big">
+        <label for="big-search">
+          <input
+            type="text"
+            id="big-search"
+            class="input-big inputs"
+            placeholder="Try &quot;Los Angeles&quot;"
+            data-qa="keypress"
+          >
+          <img
+            src="./images/Search.svg"
+            alt="icon"
+            class="big-icon"
+          >
+        </label>
+      </form>
+    </div>
+    <div class="container-bar">
+      <form action="/" method="get" data-qa="small">
+        <label for="small-search">
+          <input
+            type="text"
+            id="small-search"
+            class="input-small inputs"
+            placeholder="Try &quot;Los Angeles&quot;"
+          >
+          <img
+            src="./images/Search.svg"
+            alt="icon"
+            class="small-icon"
+          >
+        </label>
+      </form>
+    </div>
+    <!-- search bar ends here... -->
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
+.inputs {
+  box-sizing: border-box;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  width: 100%;
+  background-color: #fff;
```
meforoma, line 17: do you effectively need this?
that's a default value, so not that it impacts anything.

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  font-family: "Avenir-Book", sans-serif;
+  font-weight: 300;
+  color: #3d4e61;
+}
+
+.container-bar {
+  position: relative;
+  margin-top: 20px;
+}
```
meforoma, line 26: you could've applied this style to a `<label` element - than you wouldn't need any `<div` wrappers.
The minor difference is if you apply this style to `<label` - add display: block
```suggestion
.container-bar {
  display: block;
  position: relative;
  margin-top: 20px;
}
```

src/index.html
```diff
@@ -11,6 +11,42 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <!-- search bar starts here -->
+    <div class="container-bar">
+      <form action="/" method="get" data-qa="big">
+        <label for="big-search">
```
meforoma, line 17: take a look at the comment in `.css` file
```suggestion
    <form action="/" method="get" data-qa="big">
      <label for="big-search" class="container-bar">
```

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+
+.input-small:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.input-small:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+  line-height: 19.12px;
```
meforoma, line 79:
```suggestion
  line-height: 19px;
```
I kind of get it why you set `19.12px`, although it's not a common practice to use decimals with px...
round such to the nearest reasonable integer.
Pay attention that in figma design mock you have also a code `block` underneath `Typography`:
figma

meforoma (changes requested):
@Sardorbek1122 good job overall 👍 
just a few comments to fix and you're good.
Keep it up that way.

meforoma (approved):

PR #1267, opened 2022-03-29
Develop

alexandra-protyanova:


src/style.css
```diff
@@ -0,0 +1,102 @@
...
+  position: relative;
+  margin-top: 20px;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  background: #fff;
+  width: 100%;
+}
+
+.container1 {
```
Denchepornyuk, line 35: don't use numbers for class name

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+.image2 {
+  position: absolute;
+  top: 15px;
+  left: 13px;
+  width: 11px;
+  height: 11px;
+}
+
+.input {
+  font-family: "Avenir-book", sans-serif;
```
Denchepornyuk, line 61: use `inherit` here for font-family

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1268, opened 2022-03-29
tasks were done

vlad-demchuk:
- [DEMO LINK](https://vlad-demchuk.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://vlad-demchuk.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,85 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir.ttc);
+}
+
+@font-face {
+  font-family: "Avenir-Book";
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+.form {
+  margin: 20px 0;
+}
+
+.input {
+  width: 100%;
+  height: 70px;
+  padding-left: 62px;
+  display: inline-block;
+
+  background: #fff;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,85 @@
...
+
+.input {
+  width: 100%;
+  height: 70px;
+  padding-left: 62px;
+  display: inline-block;
+
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
```
Denchepornyuk, line 28: add `box-sizing: border-box` for selector `*`

src/style.css
```diff
@@ -0,0 +1,85 @@
...
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+
+  font-family: Avenir-Heavy, sans-serif;
```
Denchepornyuk, line 32: add font-family for html tag and here you can use `inherit`

src/style.css
```diff
@@ -0,0 +1,85 @@
...
+}
+
+.icon {
+  position: absolute;
+  top: 0;
+  left: 25px;
+}
+
+.input::placeholder {
+  font-family: Avenir, sans-serif;
```
Denchepornyuk, line 49:
```suggestion
  font-family: inherit;
```

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1280, opened 2022-03-29
Search form 1

kitaminary:


src/index.html
```diff
@@ -11,6 +11,43 @@
...
-    <h1>Search bar airbnb</h1>
+    <form
+      method="post"
+      action="#"
+      data-qa="big"
+      class="form__big"
+    >
+      <div class="search__container big__container">
+        <label>
+          <img class="img__big" src="./images/Search.svg" alt="search icon">
```
Denchepornyuk, line 22: if the HTML tag has 3 or more arguments, each must be on a new line

src/index.html
```diff
@@ -11,6 +11,43 @@
...
+
+    <form
+      method="post"
+      action="#"
+      data-qa="small"
+      class="form__small"
+    >
+      <div class="search__container small__container">
+        <label>
+          <img class="img__small" src="./images/Search.svg" alt="search icon">
```
Denchepornyuk, line 42: if the HTML tag has 3 or more arguments, each must be on a new line

src/style.css
```diff
@@ -0,0 +1,85 @@
...
+}
+
+* {
+  color: #3d4e61;
+  font-family: Avenir-book, sans-serif;
+  font-size: 16px;
+  line-height: 22px;
+  letter-spacing: 0;
+  text-align: left;
+  font-weight: 300;
```
Denchepornyuk, line 17: all these styles shouldn't be here. It's a bad idea to add these styles to the `*` selector

src/style.css
```diff
@@ -0,0 +1,85 @@
...
+}
+
+.big__container {
+  margin: 20px 0;
+}
+
+.big__search {
+  height: 70px;
+  padding-left: 62px;
+  top: 0;
```
Denchepornyuk, line 27: I believe `top: 0;` doesn't work here, because you don't have position for big__search
you can remove `top: 0`

src/style.css
```diff
@@ -0,0 +1,85 @@
...
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+}
+
+.small__search {
+  height: 42px;
+  padding-left: 33px;
+  top: 0;
```
Denchepornyuk, line 45: the same 
you can remove top: 0

Denchepornyuk (changes requested):

src/style.css
```diff
@@ -37,12 +26,17 @@
   box-sizing: border-box;
   box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
   border-radius: 4px;
+  font-family: Avenir-book, sans-serif;
```
Denchepornyuk, line 29: add `font-family` to html tag and here use `inherit`

Denchepornyuk (changes requested):

PR #1291, opened 2022-03-31
Search bar

AndriySimyachenko:
- [DEMO LINK](https://AndriySimyachenko.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://AndriySimyachenko.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,82 @@
+body {
+  font-weight: 300;
+}
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir.ttc);
+}
+
+@font-face {
+  font-family: "Avenir-Book";
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,82 @@
+body {
+  font-weight: 300;
+}
```
Denchepornyuk, line 3: add `font-weight` for html and also add `font-family`
move this selector under `@font-face`

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+.big {
+  height: 60px;
+  width: 100%;
```
Denchepornyuk, line 21: add font-family for inputs

Denchepornyuk (changes requested):
Screenshot 2022-03-31 at 12 45 34

text should bold when input is focused

PR #1294, opened 2022-03-31
First_commit_SearchBar

igoree88:


igoree88:
Не зміг пофіксити проблему з значком лупи, в меншому  SearchBar, чомусь при зменшені його розміру - не вірно відображається. Нажаль не розумію чому.

src/index.html
```diff
@@ -2,15 +2,40 @@
...
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form action="#"
+      method="post"
+      data-qa="big"
+      class="form"
+    >
```
Denchepornyuk, line 16:
```suggestion
    <form
      action="#"
      method="post"
      data-qa="big"
      class="form"
    >
```

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+}
+
+.search {
+  position: relative;
+
+}
+
+.search-field {
+  padding-left: 62px;
+  font-family: Avenir, Helvetica, sans-serif;
```
Denchepornyuk, line 27: add font-family to html and here use inherit

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  height: 70px;
+  display: block;
+  width: 100%;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+}
+
+.search::before {
```
Denchepornyuk, line 40: Screenshot 2022-04-01 at 11 14 36

try to make the icon bigger and add `background-repeat: no-repeat;`
also, center your icon using the `top` and `left` rules.

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+}
+
+.small-search {
+  position: relative;
+
+}
+
+.small-search-field {
+  padding-left: 33px;
+  font-family: Avenir, Helvetica, sans-serif;
```
Denchepornyuk, line 77: use inherit for `font-family`

src/style.css
igoree88: Thanks for the advice, I added ` background-repeat: no-repeat;`
The icon for some reason is displayed completely only in the size  like for big SearchBar (19x19px). So I made the same icon size also for small SearchBar.
I left top and left rules like in the mockup.

Denchepornyuk (approved):

src/style.css
igoree88: I fixed this problem with `background-repeat: no-repeat;`, thanks for the advice!

PR #1302, opened 2022-04-05
Search bar solution

Serhii-Vovk:
[DEMO LINK](https://Serhii-Vovk.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://Serhii-Vovk.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,77 @@
+@font-face {
+  font-family: Avenir;
+  src: url("./fonts/Avenir.ttc");
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  src: url("./fonts/Avenir-Book.ttf");
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 900;
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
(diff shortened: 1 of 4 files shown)


Bohdan-Maksymiuk (changes requested):
```
.html {
  font-weight: 300;
}
```
These styles for the font are repeated, so you can use them on the whole page and then they will be inherited (change in classes where it is needed)
But great job 🚀

src/style.css
```diff
@@ -0,0 +1,80 @@
...
+  src: url("./fonts/Avenir-Book.ttf");
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 900;
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
+* {
```
Bohdan-Maksymiuk, line 18: better for html

Bohdan-Maksymiuk (approved):

src/style.css
Serhii-Vovk: And reset to default browser settings?
* {
  all: unset;
  display: revert;
  box-sizing: border-box;
}

PR #1323, opened 2022-04-22
Develop

frontend112:
https://frontend112.github.io/layout_search-bar-airbnb/
https://frontend112.github.io/layout_search-bar-airbnb/report/html_report/


meforoma (approved):
neat 👍

PR #1372, opened 2022-05-03
2 search bars were added and stylized

Andriy3323:
- [DEMO LINK](https://andriy3323.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://andriy3323.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,99 @@
...
+  left: 26px;
+}
+
+.bigSearchField {
+  box-sizing: border-box;
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+  font-size: 16px;
+  color: #3d4e61;
+  background: white;
```
mykola-koval, line 35: if you want to change background color only, it is better to use the background-color CSS property
background is a shorthand property
https://developer.mozilla.org/en-US/docs/Web/CSS/background

mykola-koval (approved):
good job 👍

PR #1464, opened 2022-05-21
New search

Volohonskyi:


marina-tilniak (changes requested):
- fix checklist point 2 
- placeholder text differs from mockup. Fix it please

PR #1475, opened 2022-05-30
search-bar-airbnb

Ma-Anna:
https://ma-anna.github.io/layout_search-bar-airbnb/
https://ma-anna.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,71 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,71 @@
...
+  left: 13px;
+  top: 15px;
+}
+
+.searchInput {
+  font-family: Avenir, sans-serif;
+  font-size: 16px;
+  line-height: 22px;
+  width: 100%;
+  height: 70px;
```
aliceblunt8, line 40: add box-sizing here, cause now the element's width is bigger than 70px, cause the border is includes
that's why the tests is not passed

aliceblunt8 (changes requested):

src/style.css
Ma-Anna: done

PR #1507, opened 2022-06-06
was added search forms

OlgaStakhova:
- [DEMO LINK](https://OlgaStakhova.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://OlgaStakhova.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,12 +1,12 @@
 @font-face {
-  font-family: "Avenir";
-  src: url("./fonts/Avenir-Book.ttf");
+  font-family: "Avenir", sans-serif;
+  src: url("fonts/Avenir-Book.ttf");
   font-style: normal;
   font-weight: 300;
 }
 @font-face {
-  font-family: "Avenir";
-  src: url("./fonts/Avenir-Heavy.ttf");
+  font-family: "Avenir", sans-serif;
+  src: url("fonts/Avenir-Heavy.ttf");
   font-style: normal;
   font-weight: 800;
 }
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -0,0 +1,117 @@
...
+
+html {
+  box-sizing: border-box;
+}
+
+*,
+*:before,
+*:after {
+  box-sizing: inherit;
+}
```
VETALEK, line 22: 🧠

src/style.css
```diff
@@ -0,0 +1,117 @@
+@font-face {
+  font-family: "Avenir";
```
VETALEK, line 2:
```suggestion
  font-family: "Avenir", sans-serif;
```

VETALEK (changes requested):
Good 💪 
just a few improvements 😁

src/style.css
OlgaStakhova: When I added ,sans-serif in 2 and 8 line, all test was failed

VETALEK: bad 🥲
but you need the font family in the end, so don't care too much about test, if you are doing right 🙃

src/style.css
```diff
@@ -0,0 +1,119 @@
...
+
+.input-search:hover {
+  border: 1px solid rgb(225, 231, 237);
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.input-search:focus {
+  background: linear-gradient(180deg,
+  rgb(255, 255, 255) 0%, rgb(246, 246, 247) 100%);
```
VETALEK, line 115:
```suggestion
  background: linear-gradient(180deg,
    rgb(255, 255, 255) 0%,
    rgb(246, 246, 247) 100%);
```

VETALEK (approved):
Nice 🤩

PR #1521, opened 2022-06-07
Search bar was created

Ihor-Myslovskyi:
- [DEMO LINK](https://ihor-myslovskyi.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ihor-myslovskyi.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -69,13 +69,6 @@
   outline: none;
 }
 
-.bigger_form:focus,
-.smaller_form:focus {
-  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
-  border-radius: 3px;
-  outline: none;
-}
-
 .big_search::placeholder,
 .small_search::placeholder {
   font-weight: 400;


src/style.css
```diff
@@ -0,0 +1,82 @@
...
+  border-radius: 3px;
+  outline: none;
+}
+
+.bigger_form:focus,
+.smaller_form:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+  outline: none;
+}
```
VETALEK, line 77: they are the same rules 😉
and do you really need a `:focus` for forms? 😁

VETALEK (changes requested):
good 🤩 
commented 🙃

src/style.css
Ihor-Myslovskyi: I had 'too much focusin' ahah 
so sorry, already edited :)

PR #1672, opened 2022-07-12
add search bar

KseniyaGonchar:


KseniyaGonchar:
https://kseniyagonchar.github.io/layout_search-bar-airbnb/

src/index.html
```diff
@@ -11,6 +11,56 @@
...
   <body>
-    <h1>Search bar airbnb</h1>
+    <div 
+      tabindex="0" 
+      class="search-box" 
+      data-qa="big"
+    >
+      <form 
+        method="get"
+        action="/"
```
alexandra-protyanova, line 21:
```suggestion
        action="#" 
```

src/index.html
```diff
@@ -11,6 +11,56 @@
...
+      </form>
+    </div>
+    <div 
+      tabindex="0" 
+      class="search-box" 
+      data-qa="small"
+    >
+      <form 
+        method="get"
+        action="/"
```
alexandra-protyanova, line 46:
```suggestion
        action="#" 
```

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-weight: 900;
+}
+
+div[data-qa="big"] {
```
alexandra-protyanova, line 13: use classes for styles

alexandra-protyanova (changes requested):

KseniyaGonchar:
i can`t fix 2 problems(((

src/style.css
```diff
@@ -0,0 +1,144 @@
...
+  background-image: url(../src/images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 19px;
+  height: 19px;
+  top: 25px;
+  left: 26px;
+}
+
+#search-text {
```
alexandra-protyanova, line 49: don't use id for styles, use classes

alexandra-protyanova (changes requested):

src/style.css
```diff
@@ -0,0 +1,144 @@
...
+  margin: 8px;
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
+
+[data-qa="big"] {
+  width: 100%;
+  height: 70px;
+  margin: 20px auto 0;
+}
```
valerii-derkach, line 24: move this styles to class selector

src/style.css
```diff
@@ -0,0 +1,144 @@
...
+  font-weight: 300;
+  font-family: inherit;
+  color: #3d4e61;
+}
+
+[data-qa="small"] {
+  width: 100%;
+  height: 42px;
+  margin-top: 20px;
+}
```
valerii-derkach, line 94: move this styles to class selector

src/style.css
```diff
@@ -0,0 +1,144 @@
...
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  outline: none;
+  box-sizing: border-box;
+  border: 1px solid #e1e7ed;
+  font-size: 14px;
+  line-height: 19px;
+  font-weight: 300;
+  font-family: inherit;
+  padding: 0 0 0 33px;
+  border-radius: 4px;
```
valerii-derkach, line 119: if you have two or more similar elements with portions of similar styles with different values - use one of the elements as the basic case, and override necessary styles for other cases. Explanation: The point is not in the names of the classes, the point is: when there are several similar elements, ex., 2 inputs, for one we can give a class input, for example, and for the second - input input--small. We write all the styles for .input, but for .input-small we write only those styles that differ in design, and we need this second input to look a little different. Element with class .input without extra classes should also look like a full-fledged styled element.

https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md

valerii-derkach (changes requested):
please fix selectors in your code, use classes

valerii-derkach (approved):
Good job!

PR #1710, opened 2022-07-15
Search bar

olena-hapon:
https://olena-hapon.github.io/layout_search-bar-airbnb/
https://olena-hapon.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,43 +9,40 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <form 
-      action="#" 
+    <form
       class="bigSearchBar"
-      method="get"
-    > 
-      <label class="searchBar1">
-        <img 
-          src="images/Search.svg" 
-          alt="a" 
-          class="image1"
-        >
-        <input
-          type="text"
-          class="firstSearchBar"
-          placeholder="Try &quot;Los Angeles &quot"
-          data-qa="big"
-        >
-      </label>
-    </form>
-    <form 
-      action="#" 
-      class="smallSearchBar" 
-      method="get"
+      action="#"
+      method="post"
+      data-qa="big"
     >
-      <label class="searchBar2">
-        <img 
-          src="images/Search.svg" 
-          alt="a" 
-          class="image2"
-        >
-        <input
-          type="text"
-          class="secondSearchBar"
-          placeholder="Try &quot;Los Angeles &quot"
-          data-qa="small"
-        >
-      </label>
+      <img
+        src="./images/Search.svg"
+        alt="a"
+        class="imageFirst"
+      >
+      <input
+        type="text"
+        class="bigSearchInput"
+        data-qa="keypress"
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -2,15 +2,50 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form 
+      action="#" 
+      class="bigSearchBar"
+      method="get"
+    > 
+      <label class="searchBar1">
```
valerii-derkach, line 17: do not use numbers in class names

src/index.html
```diff
@@ -2,15 +2,50 @@
...
+          data-qa="big"
+        >
+      </label>
+    </form>
+    <form 
+      action="#" 
+      class="smallSearchBar" 
+      method="get"
+    >
+      <label class="searchBar2">
```
valerii-derkach, line 36: do not use numbers in class names

src/style.css
```diff
@@ -0,0 +1,107 @@
...
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.secondSearchBar:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+}
```
valerii-derkach, line 107: [STYLES] - if you have two or more similar elements with portions of similar styles with different values - use one of the elements as the basic case, and override necessary styles for other cases. Explanation: The point is not in the names of the classes, the point is: when there are several similar elements, ex., 2 inputs, for one we can give a class input, for example, and for the second - input input--small. We write all the styles for .input, but for .input-small we write only those styles that differ in design, and we need this second input to look a little different. Element with class .input without extra classes should also look like a full-fledged styled element.

valerii-derkach (changes requested):
please pass the [check list](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md)
you duplicate too much code

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+  position: relative;
+  margin: 20px 8px;
+  height: 42px;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+}
+
+.bigSearchInput {
+  font-family: Avenir, sans-serif;
```
valerii-derkach, line 40: move fonts to html tag, they will be inherited for all page,
and then you can remove them from every class style
it wold be inherited in almost all tags and classes, if not - write font-family: inherit

valerii-derkach (approved):

src/style.css
olena-hapon: Thanks! I will move.

PR #1718, opened 2022-07-19
Big and small searchbar was added

ArmPlay:
- [DEMO LINK](https://armplay.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://armplay.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -20,11 +20,11 @@ body {
   margin-top: 20px;
 }
 
-.form[data-qa="big"] {
+.form-big {
   font-size: 16px;
 }
 
-.form[data-qa="small"] {
+.form-small {
   font-size: 14px;
 }
 
@@ -69,7 +69,7 @@ body {
   --small-icon-size: 11px;
 }
 
-.search-big::before {
+.search-field::before {
   display: block;
   position: absolute;
   top: 25px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,102 @@
...
+.search-small::before {
+  display: block;
+  content: "";
+  position: absolute;
+  top: 15px;
+  left: 13px;
+  width: var(--small-icon-size);
+  height: var(--small-icon-size);
+  background: url(./images/Search.svg) center center/cover no-repeat;
+}
```
valerii-derkach, line 97: looks like you repeat the same styles in both forms,  if you have two or more similar elements with portions of similar styles with different values - use one of the elements as the basic case, and override necessary styles for other cases. Explanation: The point is not in the names of the classes, the point is: when there are several similar elements, ex., 2 inputs, for one we can give a class input, for example, and for the second - input input--small. We write all the styles for .input, but for .input-small we write only those styles that differ in design, and we need this second input to look a little different. Element with class .input without extra classes should also look like a full-fledged styled element.

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #1723, opened 2022-07-20
commit message

Irina-Kulish:


src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  line-height: 19px;
+  color: #3d4e61;
+}
+
+.input--small:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+  font-weight: bold;
+  outline: none;
+}
```
ArturErmolenko, line 70: don't repeat styles

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+.search-input {
+  position: relative;
+}
+
+.search-form {
+  margin: 20px 0;
+}
+
+input {
```
ArtemTopolskyi, line 22: do not write styles for tags, use classes

ArtemTopolskyi (approved):

PR #1782, opened 2022-08-05
solution

vmamchur:
- [DEMO LINK](https://vmamchur.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://vmamchur.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,65 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
+
+.search__form {
+  margin-top: 20px;
+  position: relative;
+}
+
+.search__input {
+  width: 100%;
+  height: 70px;
+  background-color: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px
+    rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  padding-left: 62px;
+  font-family: "Avenir", sans-serif;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,65 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
```

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+.search__input {
+  width: 100%;
+  height: 70px;
+  background-color: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px
+    rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  padding-left: 62px;
+  font-family: "Avenir", sans-serif;
```
Denchepornyuk, line 29: add font-family to html tag, here you can use inherit

Denchepornyuk (changes requested):

src/style.css
vmamchur: after I've moved font-family to html tag, 4 of the tests failed.
image
image

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  position: relative;
+}
+
+.search__input {
+  width: 100%;
+  height: 70px;
+  background-color: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px
+    rgba(61, 78, 97, 0.1);
```
Denchepornyuk, line 28: add `font-family: inherit` for inputs
Inputs can't inherit font-family from the html by default, we should add this rule for inputs

Denchepornyuk (changes requested):

src/style.css
Denchepornyuk: here is the answer)
https://github.com/mate-academy/layout_search-bar-airbnb/pull/1782#discussion_r945772294

PR #1789, opened 2022-08-07
readme.md was modified, index.html and style.css was also modified with resolved task in it

TheVovchik:
- [DEMO LINK](https://TheVovchik.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://TheVovchik.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,5 +1,6 @@
 <!doctype html>
 <html lang="en">
+
   <head>
     <meta charset="UTF-8">
     <meta
@@ -10,7 +11,49 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+    class="search"
+    action="#"
+    method="get"
+    data-qa="big"
+    >
+      <label>
+        <img
+        class="search--big__img"
+        src="./images/Search.svg"
+        alt="search"
+        >
+        <input
+        class="search__field search--big__field"
+        type="search"
+        name="search-big"
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+        >
+      </label>
+    </form>
+    <form
+    class="search"
+    action="#"
+    method="get"
+    data-qa="small"
+    >
+      <label>
+        <img
+        class="search--small__img"
+        src="./images/Search.svg"
+        alt="search"
+        >
+        <input
+        class="search__field search--small__field"
+        type="search"
+        name="search-small"
+        placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
   </body>
+
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,104 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
+
+.search {
+  margin-top: 20px;
+  position: relative;
+}
+
+.search__field {
+  width: 100%;
+  box-sizing: border-box;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -10,7 +11,49 @@
...
+        src="./images/Search.svg"
+        alt="search"
+        >
+        <input
+        class="search__field search--big__field"
+        type="search"
+        name="search-big"
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+        >
```
Denchepornyuk, line 34: fix indents for attributes, check your code with checklist

src/style.css
```diff
@@ -0,0 +1,104 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
```

src/style.css
```diff
@@ -0,0 +1,104 @@
...
+}
+
+.search {
+  margin-top: 20px;
+  position: relative;
+}
+
+.search__field {
+  width: 100%;
+  box-sizing: border-box;
```
Denchepornyuk, line 22: usually, we add box-sizing to * tag in css

src/style.css
```diff
@@ -0,0 +1,104 @@
...
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+/* #region big */
+
+.search--big__field {
+  padding-left: 62px;
+  height: 70px;
+  color: #3d4e61;
+  font-family: "Avenir", sans-serif;
```
Denchepornyuk, line 38: add font-family to html tag, here you can use inherit

src/style.css
```diff
@@ -0,0 +1,104 @@
...
+
+.search--big__field:focus {
+  outline: none;
+  font-weight: 800;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+}
+
+.search--big__field::placeholder {
+  font-family: "Avenir", sans-serif;
```
Denchepornyuk, line 60: add font-family to html tag and use inherit here

src/style.css
```diff
@@ -0,0 +1,104 @@
...
+
+/* #region small */
+
+.search--small__field {
+  padding-left: 33px;
+  height: 42px;
+  color: #3d4e61;
+  font-size: 14px;
+  line-height: 22px;
+  font-family: "Avenir", sans-serif;
```
Denchepornyuk, line 77: here the same with font-family

Denchepornyuk (changes requested):

src/style.css
```diff
@@ -1,25 +1,30 @@
...
   src: url(./fonts/Avenir-Heavy.ttf);
-  font-style: normal;
   font-weight: 800;
 }
 
+html {
+  font-family: "Avenir", sans-serif;
+}
+
+input {
```
Denchepornyuk, line 17: don't use tags in CSS, add class for inputs and use class instead of tag

Denchepornyuk (changes requested):

src/style.css
TheVovchik: [Denchepornyuk](https://github.com/Denchepornyuk) [17 hours ago](https://github.com/mate-academy/layout_search-bar-airbnb/pull/1789#discussion_r945707970)
usually, we add box-sizing to * tag in css

so I need to use tag in this case or class for adding border-box? 
I not connect resets for all styles and I first time choose to work with block in class. 
So previously I had no tag, and I add border-box to class, and you marked it like a fault, and I changed and move it to tag... and you marked tag like a fault))) what is right?)))

TheVovchik:
same in VS code        
     report | 6 Passed
     report | 0 Failed

src/index.html
```diff
@@ -10,7 +11,49 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      class="search"
+      action="#"
+      method="get"
```
Denchepornyuk, line 19: it's better to use POST method instead of GET

Denchepornyuk (approved):

src/index.html
TheVovchik: You totally right, I don't know why I used here GET (it's default)... As I know GET will show data while transfer it in page link, and POST will encrypt it. Probably I think that this data isn't sensitif...

PR #1803, opened 2022-08-10
Search bar was done

anna-borodayenko:
- [DEMO LINK](https://anna-borodayenko.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://anna-borodayenko.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,7 +10,29 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+
+    <div class="search__form search__form__big" data-qa="big">
+      <form class="search__field">
+        <input type="text"
+                class="search search__big"
+                name="searchBig"
+                placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </form>
+    </div>
+
+    <div class="search__form search__form__small" data-qa="small">
+      <form class="search__field">
+        <input type="text"
+                class="search search__small"
+                name="searchSmall"
+                placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </form>
+    </div>
+
   </body>
 </html>
+
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,77 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir.ttc);
+}
+
+* {
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+  box-sizing: border-box;
+}
+
+body {
+  margin-top: 0;
+}
+
+.search__form {
+  display: inline-block;
+  width: 100%;
+  color: #3d4e61;
+  margin: 20px auto 0;
+
+}
+
+.search__form__big {
+  line-height: 22px;
+}
+
+.search__form__small {
+  line-height: 20px;
+}
+
+.search {
+  box-sizing: border-box;
+  width: 100%;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  outline: none;
+  background-image: url("images/Search.svg");
+  background-repeat: no-repeat;
+}
+
+.search__big {
+  font-size: 16px;
+  line-height: 68px;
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -10,7 +10,29 @@
...
   <body>
-    <h1>Search bar airbnb</h1>
+
+    <div class="search__form search__form__big" data-qa="big">
+      <form class="search__field">
+        <input type="text"
+          class="search search__big"
+          name="searchBig"
+          placeholder="Try &quot;Los Angeles&quot;"
+      >
```
OleksiiBlezniuk, line 22:
```suggestion
        <input type="text"
          class="search search__big"
          name="searchBig"
          placeholder="Try &quot;Los Angeles&quot;"
        >
```

OleksiiBlezniuk: Closing bracket should have same indentation as open bracket

src/index.html
```diff
@@ -10,7 +10,29 @@
...
+      </form>
+    </div>
+
+    <div class="search__form search__form__small" data-qa="small">
```
OleksiiBlezniuk, line 26: Wow, looks like BEM 👍 

But in BEM modifiers are separated with '--', '__' is for elements and elements should have a block, but its ok, you will learn about it later 🙂

src/index.html
OleksiiBlezniuk: Fix everywhere please

src/style.css
```diff
@@ -0,0 +1,77 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir.ttc);
+}
+
+* {
```
OleksiiBlezniuk, line 6: Usually we set font styles to `html` tag, other elements will inherit it

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  font-weight: 300;
+  box-sizing: border-box;
+}
+
+body {
+  margin-top: 0;
+}
+
+.search__form {
+  display: inline-block;
```
OleksiiBlezniuk, line 17: Do you really need to set `inline-block` here? It's a block by default and I don't see any need in `inline-block` here

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  width: 100%;
+  color: #3d4e61;
+  margin: 20px auto 0;
+
+}
```
OleksiiBlezniuk, line 22:
```suggestion
  margin: 20px auto 0;
}
```

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  border-radius: 4px;
+  outline: none;
+  background-image: url("images/Search.svg");
+  background-repeat: no-repeat;
+}
+
+.search__big {
+  font-size: 16px;
+  line-height: 68px;
+  padding: 0 0 0 62px;
```
OleksiiBlezniuk, line 46: you can reset default paddings in `search` class, and set here only that padding that you need, no need to duplicate this rule in both `search__big` and `search__small`

OleksiiBlezniuk (changes requested):

src/style.css
Anna-Borodaienko: https://drive.google.com/file/d/1ugcbMyvKyc-FHUDtZvtQ6lGfUrC10LIa/view?usp=sharing
https://drive.google.com/file/d/1N-RO6yuzCaAMUdnaQ2dhls97bveg4f9A/view?usp=sharing

I don't know how it works but after moving fonts styles from tag * to tag html tests failed (
(I didn't change anything more)

OleksiiBlezniuk: Strange, it shouldn't work that way 🤔 Okay, let it be in `*` for this task

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+
+* {
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+  box-sizing: border-box;
+}
+
+body {
+  margin-top: 0;
+}
```
OleksiiBlezniuk, line 14: Why only top?

OleksiiBlezniuk (approved):

src/style.css
Anna-Borodaienko: I don't know why I use this margin at all. I deleted it.

PR #1810, opened 2022-08-11
solution

artem-nasadchuk:


src/style.css
```diff
@@ -0,0 +1,77 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
```

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+.search__field {
+  width: 100%;
+  height: 70px;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  padding-left: 62px;
+  font-family: "Avenir", sans-serif;
```
Denchepornyuk, line 28: add font-family to html tag, here you can use inherit

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+.icon--small {
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
+
+.search__field--small::placeholder {
+  font-family: "Avenir", sans-serif;
```
Denchepornyuk, line 62: and font-family to html and here you can use inherit

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1836, opened 2022-08-15
Solution

anton-iskryk:
[https://anton-iskryk.github.io/layout_search-bar-airbnb/](url)
[https://anton-iskryk.github.io/layout_search-bar-airbnb/report/html_report/](url)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -12,10 +12,6 @@
   font-style: normal;
 }
 
-body {
-  box-sizing: border-box;
-}
-
 .form {
   position: relative;
   margin-top: 20px;
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,83 @@
+@font-face {
+  font-family: "Avenir";
+  font-weight: 300;
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 800;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  font-style: normal;
+}
+
+.form {
+  position: relative;
+  margin-top: 20px;
+}
+
+.searchBar {
+  font-family: "Avenir", sans-serif;
```
Denchepornyuk, line 21: add `font-family` and all default styles for text to `html` tag

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+
+  font-size: 14px;
+}
+
+.searchBar:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.searchBar::placeholder {
+  font-family: "Avenir", sans-serif;
```
Denchepornyuk, line 48: you can use here `inherit` you already added font-family style

Denchepornyuk (changes requested):
good job 👍

src/style.css
anton-iskryk: Should I use 'inherit' for font-size and font-weight in all other cases like in .searchBar and .searchBar::placeholder?

Denchepornyuk (approved):

PR #1847, opened 2022-08-15
added two search bars

vladimir-dublya:
- [DEMO LINK](https://vladimir-dublya.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://vladimir-dublya.github.io/layout_search-bar-airbnb/report/html_report/)


.linthtmlrc.json
```diff
@@ -1,3 +1,49 @@
 {
-  "extends": "@mate-academy/linthtml-config"
+  "attr-bans": [
+    "align",
+    "background",
+    "bgcolor",
```
OleksiiBlezniuk, line 5: Please do not add config or package files to future PRs

src/index.html
```diff
@@ -11,6 +11,42 @@
...
+      action="/"
+      method="get"
+      data-qa="big"
+      class="search-form"
+    >
+      <input
+      type="text"
+      name="x1"
+      id="x"
+      placeholder="Try &quot;Los Angeles&quot;">
```
OleksiiBlezniuk, line 25:
```suggestion
      <input
        type="text"
        name="x1"
        id="x"
        placeholder="Try &quot;Los Angeles&quot;"
      >
```

OleksiiBlezniuk: What is x? And what is x1? Please use more clear names

src/index.html
```diff
@@ -11,6 +11,42 @@
...
+      <img
+        class="icon"
+        src="images/Search.svg"
+        alt="search"
+      >
+    </form>
+
+    <form
+      action="/"
+      method="get"
```
OleksiiBlezniuk, line 35: it's a default value

src/index.html
```diff
@@ -11,6 +11,42 @@
...
+      data-qa="small"
+      class="search-form"
+    >
+      <input
+      type="text"
+      name="y1"
+      id="y"
+      placeholder="Try &quot;Los Angeles&quot;">
```
OleksiiBlezniuk, line 43:
```suggestion
      <input
        type="text"
        name="y1"
        id="y"
        placeholder="Try &quot;Los Angeles&quot;"
      >
```

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+  font-style: normal;
+  font-weight: 300;
+  src: url(fonts/Avenir-Book.ttf);
+}
+@font-face {
+  font-family: Avenir;
+  font-style: normal;
+  font-weight: 800;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
```
OleksiiBlezniuk, line 38: Usually all imports are on the top of CSS file

src/style.css
```diff
@@ -0,0 +1,97 @@
+#x {
```
OleksiiBlezniuk, line 1: Why not class?

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#y:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+  font-weight: 800;
+}
+
+#x::placeholder {
```
OleksiiBlezniuk, line 84: Most of this styles are duplicated, there is no need to do it.
Create class for main input (for example big), then set only values that should be different for second class (for example for small input). Then just use two classes on small input

This will help to reduce your styles

src/style.css
```diff
@@ -0,0 +1,97 @@
+#x {
+  box-sizing: border-box;
+  font-family: Avenir, sans-serif;
```
OleksiiBlezniuk, line 3: Set font styles to html tag, other elements will inherit (except placeholders, you have to use `inherit` for them)

OleksiiBlezniuk (changes requested):
DRY, use classes for styles

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+  font-weight: 800;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+html {
+  font-family: Avenir, sans-serif;
+}
+
+.form {
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 19: usually we set this rule for every elements using * selector

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  color: #3d4e61;
+  width: 100%;
+  font-family: Avenir, sans-serif;
+}
+
+.bigForm {
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
```
OleksiiBlezniuk, line 29: it is same as in smallForm, move it to form, no need to duplicate it

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+}
+
+.form {
+  box-sizing: border-box;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  color: #3d4e61;
+  width: 100%;
+  font-family: Avenir, sans-serif;
```
OleksiiBlezniuk, line 25: use `inherit`

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+.bigForm:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.bigForm:focus {
+  font-weight: 800;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+  outline: none;
+}
```
OleksiiBlezniuk, line 45: Just add it to `form` class, both input have same styles in `focus` state

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+
+.icon1 {
+  position: absolute;
+  top: 15px;
+  left: 13px;
+  width: 11px;
+}
+
+.smallForm:hover {
+  background: #fff;
```
OleksiiBlezniuk, line 70: it is already set in form

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+.bigForm::placeholder {
+  font-weight: 300;
+  color: #3d4e61;
+}
+
+.smallForm::placeholder {
+  font-weight: 300;
+  color: #3d4e61;
+}
+
```
OleksiiBlezniuk, line 88: just add it to `form`, don't duplicate it

OleksiiBlezniuk (changes requested):

src/index.html
OleksiiBlezniuk: It means that there is no need to set it here, it can be removed

src/style.css
```diff
@@ -0,0 +1,80 @@
...
+.icon {
+  position: absolute;
+  top: 25px;
+  left: 26px;
+}
+
+.smallForm {
+  display: block;
```
OleksiiBlezniuk, line 58: Input is already inline-block, why do you need it?

OleksiiBlezniuk (approved):

PR #1850, opened 2022-08-15
layout_search-bar-airbnb

xiimiik:
- [DEMO LINK](https://xiimiik.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://xiimiik.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,4 +1,4 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="UTF-8">
@@ -11,6 +11,45 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="/"
+      data-qa="big"
+      method="get"
+      class="search-form"
+    >
+      <input
+        type="text"
+        name="bigSearchBar"
+        data-qa="keypress"
+        class="searchBar"
+        placeholder="Try &#34;Los Angeles&#34;"
+      >
+
+      <img
+        class="image"
+        src="images/Search.svg"
+        alt="search"
+      >
+    </form>
+
+    <form
+    action="/"
+    data-qa="small"
+    method="get"
+    class="search-form"
+    >
+      <input
+        type="text"
+        name="smallSearchBar"
+        class="searchBar smallSearchBar"
+        placeholder="Try &#34;Los Angeles&#34;"
+      >
+
+      <img
+        class="imageSmall"
+        src="images/Search.svg"
+        alt="search"
+      >
+    </form>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -33,10 +33,10 @@
     </form>
 
     <form
-    action="/"
-    data-qa="small"
-    method="get"
-    class="search-form"
+      action="/"
+      data-qa="small"
+      method="get"
+      class="search-form"
     >
       <input
         type="text"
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -11,6 +11,45 @@
...
+        alt="search"
+      >
+    </form>
+
+    <form
+    action="/"
+    data-qa="small"
+    method="get"
+    class="search-form"
+    >
```
OleksiiBlezniuk, line 40:
```suggestion
    <form
      action="/"
      data-qa="small"
      method="get"
      class="search-form"
    >
```

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.search-form {
+  margin-top: 20px;
+  position: relative;
+}
+
+.searchBar {
+  box-sizing: border-box;
+  font-family: Avenir, sans-serif;
```
OleksiiBlezniuk, line 22: Set font styles in html tag, other elements should inherit it

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  top: 25px;
+  left: 25px;
+}
+
+.imageSmall {
+  position: absolute;
+  top: 15px;
+  left: 13px;
+  width: 11px;
+}
```
OleksiiBlezniuk, line 53: This two classes has duplicated styles, create on class for big image and one for small (where you will only overwrite size values) then add two classes to small image

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  background-color: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+  border: 1px solid #e1e7ed;
+  outline: none;
+}
+
+.searchBar::placeholder,
+.smallSearchBar::placeholder {
+  font-weight: 300;
+}
```
OleksiiBlezniuk, line 70: just add it to `searchBar` class, there is no need to duplicate it

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+
+.searchBar:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.searchBar:focus {
+  font-weight: 900;
+  background-color: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+  border: 1px solid #e1e7ed;
```
OleksiiBlezniuk, line 63: You already add this style, no need to duplicate

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  width: 11px;
+}
+
+.searchBar:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.searchBar:focus {
+  font-weight: 900;
+  background-color: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
OleksiiBlezniuk, line 61: gradient can't be used as color, it should be just a background

OleksiiBlezniuk (changes requested):
Good job, left few comments to fix 🙂

src/index.html
```diff
@@ -11,6 +11,45 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="/"
+      data-qa="big"
+      method="get"
```
OleksiiBlezniuk, line 17: It's a default value, no need to set it

src/style.css
OleksiiBlezniuk: This border can be removed since you've set same border to `searchBar` class

OleksiiBlezniuk (approved):

PR #1876, opened 2022-08-15
layout_search-bar-airbnb

SanyaKovalenko:


src/style.css
```diff
@@ -0,0 +1,64 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-variant: normal;
```

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  font-weight: bold;
+}
+
+label {
+  position: relative;
+  display: flex;
+  margin-top: 20px;
+}
+
+label img {
```
Denchepornyuk, line 19: don't use tags in css, add classes in html and use classes here

Denchepornyuk (changes requested):
Screenshot 2022-08-16 at 13 27 02
Where are links for PR description?

SanyaKovalenko:
Done

Denchepornyuk (approved):

PR #1883, opened 2022-08-16
Creating search bar airbnbn

avkovlit:
[DEMO LINK](https://avkovlit.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://avkovlit.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,7 +10,48 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="/"
+      method="get"
+      class="searchBarForm"
+      data-qa="big"
+    >
+      <input
+        type="text"
+        name="bigSearchBar"
+        placeholder="Try &#34;Los Angeles&#34;"
+        data-qa="keypress"
+        class="searchBarBig"
+      >
+      <img
+        class="bigImage"
+        src="./images/Search.svg"
+        alt="searchImage"
+      >
+    </form>
+
+    <form
+      action="/"
+      method="get"
+      class="searchBarForm"
+      data-qa="small"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -10,7 +10,48 @@
...
+        name="bigSearchBar"
+        placeholder="Try &#34;Los Angeles&#34;"
+        data-qa="keypress"
+        class="searchBarBig"
+      >
+      <img
+        class="bigImage"
+        src="./images/Search.svg"
+        alt="searchImage"
+      >
```
Denchepornyuk, line 32: let's add label for input and add img to the label

src/index.html
```diff
@@ -10,7 +10,48 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="/"
+      method="get"
```
Denchepornyuk, line 17: use POST method instead of GET

src/style.css
```diff
@@ -0,0 +1,71 @@
...
+
+.searchBarForm {
+  margin-top: 20px;
+  position: relative;
+
+}
+
+.searchBarBig {
+  box-sizing: border-box;
+  font-family: Avenir, sans-serif;
```
Denchepornyuk, line 23: add font-family to html and use inherit for inputs

Denchepornyuk (changes requested):

src/style.css
avkovlit: tests don't work with html tag 
CleanShot 2022-08-16 at 15 09 14@2x

I can use * { } and tests will be pass, but I think it's not good idea

Denchepornyuk (approved):

PR #1897, opened 2022-08-16
search bars

AniaHna:
[DEMO LINK](https://AniaHna.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://AniaHna.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,25 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <div data-qa="big" class="bar big">
+      <form action="#" method="get">
+        <input
+          class="inp bi"
+          data-qa="keypress"
+          type="text"
+          placeholder="Try  &quot;Los Angeles&quot;"
+        >
+      </form>
+    </div>
+    <div data-qa="small" class="bar small">
+      <form action="#" method="get">
+        <input
+          class="inp sm"
+          data-qa="keypress"
+          type="text"
+          placeholder="Try  &quot;Los Angeles&quot;"
+        >
+      </form>
+    </div>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,87 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-weight: normal;
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+
+}
+
+.bi {
+  width: calc(100% - 62px);
+  height: 70px;
+  border-radius: 4px;
+  border: solid 1px #e1e7ed;
+  padding-left: 62px;
+  margin: 20px auto;
+  box-shadow: 0 1px 8px;
+}
+
+.big {
+  position: relative;
+  width: 100%;
+}
+
+.small {
+  position: relative;
+  width: 100%;
+
+}
+
+.sm {
+  width: calc(100% - 33px);
+  height: 42px;
+  border-radius: 4px;
+  border: solid 1px #e1e7ed;
+  padding-left: 33px;
+  box-shadow: 0 1px 8px;
+}
+
+.big::before {
+  content: url(./images/Search.svg);
+  position: absolute;
+  width: 19px;
+  height: 19px;
+  z-index: 2;
+  top: 50%;
+  left: 26px;
+  transform: translateY(-50%);
+}
+
+.small::before {
+  content: " ";
+  position: absolute;
+  background-image: url(./images/Search.svg);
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,25 @@
...
+          data-qa="keypress"
+          type="text"
+          placeholder="Try  &quot;Los Angeles&quot;"
+        >
+      </form>
+    </div>
+    <div data-qa="small" class="bar small">
+      <form action="#" method="get">
+        <input
+          class="inp sm"
```
OleksiiBlezniuk, line 27: It's not a good idea to short class names that way, it's hard to understand what it means

src/style.css
```diff
@@ -0,0 +1,87 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-weight: normal;
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+
+}
```
OleksiiBlezniuk, line 10:
```suggestion
html {
  font-family: "Avenir", sans-serif;
}
```

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+.big {
+  position: relative;
+  width: 100%;
+}
+
+.small {
+  position: relative;
+  width: 100%;
+
+}
```
OleksiiBlezniuk, line 31: It's a same styles, don't repeat yourself, just use one class

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  font-weight: normal;
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+
+}
+
+.bi {
+  width: calc(100% - 62px);
```
OleksiiBlezniuk, line 13: Strange solution, instead of width calculating you can set paddings to parent container

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  height: 70px;
+  border-radius: 4px;
+  border: solid 1px #e1e7ed;
+  padding-left: 62px;
+  margin: 20px auto;
+  box-shadow: 0 1px 8px;
+}
```
OleksiiBlezniuk, line 20: Some of this styles are same as in `sm` class, DRY, apply them to `inp` class, and override sizes in specific for size classes

OleksiiBlezniuk (changes requested):
Please do not ignore tests, I believe that you can fix most of them
Also you markup is a bit different from mockup, fix it please

Your markup
Screenshot 2022-08-17 at 11 15 03

Mockup
Screenshot 2022-08-17 at 11 15 36

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  padding-right: 33px;
+}
+
+.input {
+  border: 1px solid #e1e7ed;
+  width: 100%;
+  border-radius: 4px;
+  color: #3d4e61;
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  font-family: Avenir, sans-serif;
```
OleksiiBlezniuk, line 38: set font styles to `html` tag

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  font-weight: 300;
+}
+
+.input:hover {
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+}
+
+.input:focus {
+  border-radius: 3px;
```
OleksiiBlezniuk, line 47: It's a too small change, it's strange that it is in figma, but usually we don't change border-radius to 1px on state change. I think it can be removed

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  margin-top: 20px;
+}
+
+.big {
+  padding-right: 62px;
+}
+
+.small {
+  padding-right: 33px;
+}
```
OleksiiBlezniuk, line 30: Wait, why do you need this paddings? Your input should take all free width from page

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  font-size: 14px;
+}
+
+.bar::before {
+  content: " ";
+  position: absolute;
+  background-image: url(./images/Search.svg);
+  top: 50%;
+  transform: translateY(-50%);
+  z-index: 2;
```
OleksiiBlezniuk, line 74: It's unnecessary here, it will work same way if we'll remove it

OleksiiBlezniuk (changes requested):
Please fix your inputs widths

Screenshot 2022-08-17 at 17 49 15

Also it will be better if you'll make your commit messages more informative, better to shortly  describe changes that you've made

OleksiiBlezniuk (approved):
Good job!

PR #1898, opened 2022-08-16
Develop

Ihor-1993:
- [DEMO LINK](https://Ihor-1993.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Ihor-1993.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,11 +13,11 @@
 
   <body>
     <form
-    class="first_form"
-    action="url"
-    method="post"
-    data-qa="big"
-  >
+      class="first_form"
+      action="url"
+      method="post"
+      data-qa="big"
+    >
 
       <label>
         <img
--- a/src/index.html
+++ b/src/index.html
@@ -10,7 +10,52 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+
+    <form
+      class="first_form"
+      action="url"
+      method="post"
+      data-qa="big"
+    >
+
+      <label>
+        <img
+          class="bigLoop"
+          src="images/Search.svg"
+          alt="loop"
+        >
+
+        <input
+          type="search"
+          data-qa="keypress"
+          class="bigSearchBar"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
+
+    <form
+      class="second_form"
+      data-qa="small"
+      action="url"
+      method="post"
+    >
+      <label>
+        <img
+          class="smallLoop"
+          src="images/Search.svg"
+          alt="loop"
+        >
+
+        <input
+          type="search"
+          class="smallSearchBar"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,148 @@
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 700;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+.bigSearchBar::placeholder {
+  font-family: Avenir, Arial, Helvetica, sans-serif;
+  font-size: 16px;
+}
+
+.smallSearchBar::placeholder {
+  font-family: Avenir, Arial, Helvetica, sans-serif;
+  font-size: 14px;
+}
+
+.first_form {
+  margin-top: 20px;
+  position: relative;
+}
+
+.second_form {
+  margin-top: 20px;
+  position: relative;
+}
+
+.bigSearchBar {
+  width: 100%;
+  border: 1px solid #e1e7ed;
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -10,7 +10,51 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      class="first_form"
+      action="url"
```
OleksiiBlezniuk, line 17:
```suggestion
      action="#"
```

OleksiiBlezniuk: action - is a link where your form will send results, and "url" is not looks like real link 🙂

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 700;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+.bigSearchBar::placeholder {
+  font-family: Avenir, Arial, Helvetica, sans-serif;
```
OleksiiBlezniuk, line 14: Usually we set font styles to `html` tag, then almost all elements will inherit it (if not, you can set `inherit` yourself)

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+
+.first_form {
+  margin-top: 20px;
+  position: relative;
+}
+
+.second_form {
+  margin-top: 20px;
+  position: relative;
+}
```
OleksiiBlezniuk, line 31: Are there any differences between this styles? Use one class for both forms, don't repeat yourself

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+.smallSearchBar {
+  width: 100%;
+  border: 1px solid #e1e7ed;
+  background: #fff;
+  box-shadow: 0 1px 8px rgba(61,78,97,0.1);
+  border-radius: 4px;
+  padding-bottom: 2px;
+  height: 42px;
+  padding-left: 33px;
+}
```
OleksiiBlezniuk, line 52: Same here, just apply big searchbar class to small and override sizes styles for small search bar, don't repeat yourself

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+  left: 26px;
+  top: 25px;
+}
+
+.smallLoop {
+  position: absolute;
+  left: 13px;
+  top: 15px;
+  height: 11px;
+  width: 11px;
```
OleksiiBlezniuk, line 65: Same here with position property

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+}
+
+.bigSearchBar:hover {
+  box-shadow: 0 3px 8px rgba(61,78,97,0.2);
+}
+
+.smallSearchBar:hover {
+  box-shadow: 0 3px 8px rgba(61,78,97,0.2);
+}
```
OleksiiBlezniuk, line 74: DRY

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+  font-weight: 700;
+}
+
+.smallSearchBar:focus {
+  background: linear-gradient(180deg,#fff,#f6f6f7);
+  border-radius: 3px;
+  color: #3d4e61;
+  font-style: normal;
+  font-weight: 800;
+}
```
OleksiiBlezniuk, line 89: DRY

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+
+.search-form {
+  margin-top: 20px;
+  position: relative;
+}
```
OleksiiBlezniuk, line 94: Use this class for forms instead of `first_form` and `second_form`

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+  padding: 23px 62px;
+  font-size: 16px;
+  line-height: 22px;
+}
+
+.search-bar-small {
+  padding: 11px 33px 10px;
+  font-size: 14px;
+  line-height: 19px;
+}
```
OleksiiBlezniuk, line 116: This is right way to do it
I don't understand why you've written right styles here, but used duplicated classes instead of this

OleksiiBlezniuk (changes requested):
Remove duplicated styles

src/style.css
```diff
@@ -0,0 +1,74 @@
...
+.bigSearchBar,
+.smallSearchBar {
+  width: 100%;
+  border: 1px solid #e1e7ed;
+  background: #fff;
+  box-shadow: 0 1px 8px rgba(61,78,97,0.1);
+  border-radius: 4px;
+  height: 70px;
+  padding-left: 62px;
+}
```
OleksiiBlezniuk, line 37: I meant apply two classes in html, not same styles for both classes

src/style.css
```diff
@@ -0,0 +1,74 @@
...
+  box-shadow: 0 3px 8px rgba(61,78,97,0.2);
+}
+
+.bigSearchBar:focus,
+.smallSearchBar:focus {
+  background: linear-gradient(180deg,#fff,#f6f6f7);
+  border-radius: 3px;
+  color: #3d4e61;
+  font-weight: 700;
+}
```
OleksiiBlezniuk, line 70: Again, if you'll apply two classes to small input, there will be no need to duplicate this rules

OleksiiBlezniuk (approved):

PR #1915, opened 2022-08-17
solution-v1: initial

igor-1987:
- [DEMO LINK](https://igor-1987.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://igor-1987.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,9 +8,47 @@
     >
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Search bar airbnb</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
   </head>
-  <body>
-    <h1>Search bar airbnb</h1>
+  <body class="body">
+    <form
+      method="get"
+      class="form"
+      action="/"
+      data-qa="big"
+    >
+      <img
+        src="./images/Search.svg"
+        alt="Search icon large"
+        class="img img-lg"
+      >
+      <input
+        class="search search-lg"
+        type="search"
+        name="searchLg"
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+      >
+    </form>
+
+    <form
+      method="get"
+      action="/"
+      data-qa="small"
+      class="form"
+    >
+      <img
+        src="./images/Search.svg"
+        alt="Search icon small"
+        class="img img-sm"
+      >
+      <input
+        type="search"
+        name="searchSm"
+        placeholder="Try &quot;Los Angeles&quot;"
+        class="search search-sm"
+        data-qa="keypress"
+      >
+    </form>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,9 +8,47 @@
...
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Search bar airbnb</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
   </head>
-  <body>
-    <h1>Search bar airbnb</h1>
+  <body class="body">
+    <form
+      method="get"
```
OleksiiBlezniuk, line 15: it's a default value, so this attribute is unnecessary here

src/style.css
```diff
@@ -0,0 +1,89 @@
...
+  font-size: 12px;
+  line-height: 19px;
+}
+
+.search::placeholder {
+  color: #3d4e61;
+}
+
+.search-lg::placeholder {
+  font-size: 16px;
```
OleksiiBlezniuk, line 73: better to use `inherit` here

src/style.css
```diff
@@ -0,0 +1,89 @@
...
+  font-size: 14px;
+}
+
+.search:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
```
OleksiiBlezniuk, line 86: This border was already set in `search` class, so there is no need to duplicate this value

src/style.css
```diff
@@ -0,0 +1,89 @@
...
+  border-radius: 3px;
```
OleksiiBlezniuk, line 87: This is too minor change, and could be skipped

src/style.css
```diff
@@ -0,0 +1,89 @@
...
+
+.search-sm::placeholder {
+  font-size: 14px;
+}
+
+.search:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
```
OleksiiBlezniuk, line 84: Seems that you forgot to disable outline 🙂

src/style.css
```diff
@@ -0,0 +1,89 @@
...
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
+
+.search {
+  width: 100%;
+  color: #3d4e61;
+  font-family: inherit;
+  font-weight: 300;
```
OleksiiBlezniuk, line 47: inherit

src/style.css
OleksiiBlezniuk: Screenshot 2022-08-18 at 10 27 08

OleksiiBlezniuk (changes requested):
GJ, but there are some small comments to fix

src/style.css
igor-1987: Thank you for review! I guess work with pixels was little bit painful so I've made couple mistakes after that. All of remarks make sense.

OleksiiBlezniuk (approved):
GJ!

PR #1936, opened 2022-08-23
added two search bars

katya-suvorova:
https://katya-suvorova.github.io/layout_search-bar-airbnb/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,48 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form 
+      action="#" 
+      method="get"
+      data-qa="big"
+      class="form_search" 
+    >
+      <img 
+        src="images/Search.svg" 
+        alt="icon"
+        class="icon"
+      >
+      <label for="s">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,48 @@
...
+      class="form_search" 
+    >
+      <img 
+        src="images/Search.svg" 
+        alt="icon"
+        class="icon"
+      >
+      <label for="s">
+        <input
+          id="s"
```
Bohdan-Maksymiuk, line 27: id should be a little bit more unique and than `s`

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  width: 19px;
+  height: 19px;
+}
+
+.icon_2 {
+  top: 15px;
+  left: 13px;
+  width: 11px;
+  height: 11px;
+}
```
Bohdan-Maksymiuk, line 34: avoid numbers, `first`, `second` in names. It is not clear.
Use `--small` modifier instead

Bohdan-Maksymiuk (changes requested):

marina-tilniak (approved):

PR #1940, opened 2022-08-24
search bar

metinbicaksiz:
- [DEMO LINK](https://metinbicaksiz.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://metinbicaksiz.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,121 @@
...
+
+.input:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.input-big {
+  padding-left: 62px;
+  font-style: 16px;
+  line-height: 22px;
+  align-self: center;
```
amitpatiljc, line 93: i think you should remove ` align-self: center;` here and it should be okay!

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+
+.input {
+  width: 100%;
+  height: 100%;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  color: #3d4e61;
+  font-weight: 300;
+  font-family: inherit;
+  box-shadow: 0 1px 8px rgb(61 78 97 / 10%) ;
```
amitpatiljc, line 81: have this as mentioned in figma
` box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);`

amitpatiljc (changes requested):
Good job , just consider mentioned changes and i think you are good to go!

src/style.css
metinbicaksiz: I did it, however, it didnt solve the problem.

src/style.css
metinbicaksiz: I think what you proposed is for the hover state and when I apply what you suggested, I fail 3 tests out of 6. With mine, I only fail 1 of them :D 
but still not all six :(

amitpatiljc (approved):
Good Job!!

PR #1950, opened 2022-08-27
Search was added

a-kulaev:
https://a-kulaev.github.io/layout_search-bar-airbnb/
https://a-kulaev.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,99 @@
+@font-face {
+  font-family: "Avenir";
+  font-weight: 400;
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 900;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+* {
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+  box-sizing: border-box;
+}
+
+.form {
+  margin-top: 20px;
+}
+
+.label {
+  display: block;
+  position: relative;
+}
+
+.labelBig::before {
+  content: "";
+  position: absolute;
+  width: 19px;
+  height: 19px;
+  top: 25px;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,99 @@
...
+  font-family: "Avenir";
+  font-weight: 900;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+* {
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+  box-sizing: border-box;
+}
```
SanyaBratashchuk, line 17: Don't apply any styles except box-sizing to the * selector, you can apply font-family and font-weight to the html tag

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+.search:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  box-shadow: none;
+  font-weight: bold;
+  outline: none;
+}
+
+.bigSearch {
+  height: 70px;
+  padding: 0 17px 0 62px;
```
SanyaBratashchuk, line 81: I guess if you used verical paddings you could avoid hardcoding the heigth

SanyaBratashchuk (changes requested):

src/style.css
a-kulaev: I made changes but now I fail github checks. Although npm test works fine on my computer, github tells me that all checks have failed. Can You explain me why?

src/style.css
a-kulaev: > I guess if you used verical paddings you could avoid hardcoding the heigth

I made changes but now I fail github checks. Although npm test works fine on my computer, github tells me that all checks have failed. Can You explain me why?

SanyaBratashchuk (approved):

PR #1967, opened 2022-09-01
search bars

Tetiana-Kupchynska:
[DEMO LINK](https://tetiana-kupchynska.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://tetiana-kupchynska.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,65 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 800;
+}
+
+.search-form {
+  position: relative;
+  margin-top: 20px;
+}
+
+.image {
+  position: absolute;
+  width: 19px;
+  height: 19px;
+  left: 26px;
+  top: 25px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  font-weight: 800;
+}
+
+.search-form {
+  position: relative;
+  margin-top: 20px;
+}
+
+.image {
+  position: absolute;
```
SanyaBratashchuk, line 19: to avoid duplication create class image and apply to it position absolute, then create to additional classes image--big, image--small and apply other rules to them

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.field {
+  box-sizing: border-box;
+  height: 70px;
+  width: 100%;
+  line-height: 22px;
+  padding-left: 62px;
+  font-size: 16px;
+  font-family: Avenir, sans-serif;
```
SanyaBratashchuk, line 41: move it to the `html` tag

SanyaBratashchuk (changes requested):

src/style.css
Tetiana-Kupchynska: while moving it to the html tag, instead of html rules, default input styles are applied and tests fail

Screenshot

PR #1983, opened 2022-09-05
Search bars

BeginerAlex:
Текст в Placeholder немного не подходит по тесту

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,72 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+@font-face {
+  font-family: "Avenir";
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,72 @@
...
+}
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: bold;
+}
+
+.input {
+  font-family: "Avenir", sans-serif;
```
liliia-livandovskaya, line 15: let's move font styles to the `html` selector and inherit it where needed

liliia-livandovskaya (changes requested):
GJ 👍  Read about box-sizing: border-box and use it. Also pay attention to comments

marina-tilniak (approved):

PR #2043, opened 2022-09-20
Develop

nazarl1997:


src/style.css
```diff
@@ -0,0 +1,83 @@
...
+  position: absolute;
+  top: 50%;
+  transform: translate(0, -50%);
+  width: 11px;
+}
+
+.search_bar-1::placeholder,
+.search_bar-2::placeholder {
+  color: #3d4e61;
+  font-family: "Avenir", serif;
```
alexandra-protyanova, line 81: move to `html` selector and use `inherit` where needed

alexandra-protyanova (changes requested):
Let us remove default outline from input and it's possible to fix tests, looks like your centering  doesn't work as it have to

marina-tilniak (approved):
outline варто прибрати, так як у тебе є дизайн для інпута у фокусі
image

PR #2091, opened 2022-09-28
added content

andrii-bassarab:
- [DEMO LINK](https://andrii-bassarab.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://andrii-bassarab.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,73 @@
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  src: url(fonts/Avenir-Book.ttf);
+}
```
denys-danyliuk, line 5: add second font as well

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+  border-radius: 4px;
+}
+
+.big_input {
+  font-weight: 300;
+  padding-left: 62px;
+  height: 70px;
+  font-size: 16px;
+  line-height: 22px;
+}
```
denys-danyliuk, line 34: move this styles to input

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+
+.small_input {
```
denys-danyliuk, line 36:
```suggestion
.input--small {
```
2 point in checklist

src/style.css
denys-danyliuk: 2 point in checklist

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+
+.icon {
+  position: absolute;
+  padding-left: 26px;
+  padding-top: 25px;
+  width: 19px;
+  height: 19px;
+}
+
+.small_icon {
```
denys-danyliuk, line 52:
```suggestion
.icon--small {
```

src/style.css
denys-danyliuk: 7 point in checklist

denys-danyliuk (changes requested):

src/style.css
andrii-bassarab: could you please write why it is better to use "--" than "_" in the class name

denys-danyliuk: because this is a modifier, you will learn more about it when you will start using BEM

denys-danyliuk (approved):
🚀

PR #2107, opened 2022-09-28
Layout search bar 2

Alessandranitti15:
https://Alessandranitti15.github.io/layout_search-bar-airbnb/
https://Alessandranitti15.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,9 +11,17 @@
   </head>
 
   <body>
-    <form action="/" method="get" data-qa="big">
+    <form
+    action="/"
+    method="get"
+    data-qa="big"
+    >
       <label class="wrapper">
-        <img class="search-icon1" src="images/Search.Svg" alt="seach lens">
+        <img
+        class="search-icon1"
+        src="images/Search.Svg"
+        alt="seach lens"
+        >
         <input
         class="searchBig"
         placeholder="Try &quot;Los Angeles&quot;"
@@ -23,9 +31,17 @@
       </label>
     </form>
 
-    <form action="/" method="get" data-qa="small">
+    <form
+    action="/"
+    method="get"
+    data-qa="small"
+    >
       <label class="wrapper">
-        <img class="search-icon2" src="images/Search.Svg" alt="seach lens">
+        <img
+        class="search-icon2"
+        src="images/Search.Svg"
+        alt="seach lens"
+        >
         <input
         class="searchSmall"
         placeholder="Try &quot;Los Angeles&quot;"
--- a/src/index.html
+++ b/src/index.html
@@ -12,18 +12,20 @@
 
   <body>
     <form
-    action="/"
+    class="search-form"
     method="get"
+    action="/"
     data-qa="big"
     >
       <label class="wrapper">
         <img
-        class="search-icon1"
+        class="icon search-icon1"
         src="images/Search.Svg"
         alt="search lens"
         >
         <input
         class="searchBig searchBar"
+        name="search-big"
         placeholder="Try &quot;Los Angeles&quot;"
         type="text"
         data-qa="keypress"
@@ -32,18 +34,20 @@
     </form>
 
     <form
+    class="search-form"
     action="/"
     method="get"
     data-qa="small"
     >
       <label class="wrapper">
         <img
-        class="search-icon2"
+        class="icon search-icon2"
         src="images/Search.Svg"
         alt="seach lens"
         >
         <input
         class="searchSmall searchBar"
+        name="search-small"
         placeholder="Try &quot;Los Angeles&quot;"
         type="text"
         >
--- a/src/index.html
+++ b/src/index.html
@@ -12,44 +12,44 @@
 
   <body>
     <form
-    class="search-form"
-    method="get"
-    action="/"
-    data-qa="big"
+      class="search-form"
+      method="get"
+      action="/"
+      data-qa="big"
     >
       <label class="wrapper">
         <img
-        class="icon search-icon1"
-        src="images/Search.Svg"
-        alt="search lens"
+          class="icon"
+          src="images/Search.Svg"
+          alt="search lens"
         >
         <input
-        class="searchBig searchBar"
-        name="search-big"
-        placeholder="Try &quot;Los Angeles&quot;"
-        type="text"
-        data-qa="keypress"
+          class="search"
+          name="search-big"
+          placeholder="Try &quot;Los Angeles&quot;"
+          type="text"
+          data-qa="keypress"
         >
       </label>
     </form>
 
     <form
-    class="search-form"
-    action="/"
-    method="get"
-    data-qa="small"
+      class="search-form"
+      action="/"
+      method="get"
+      data-qa="small"
     >
       <label class="wrapper">
         <img
-        class="icon search-icon2"
-        src="images/Search.Svg"
-        alt="seach lens"
+          class="icon search-icon2"
+          src="images/Search.Svg"
+          alt="seach lens"
         >
         <input
-        class="searchSmall searchBar"
-        name="search-small"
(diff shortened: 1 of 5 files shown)


jstmpelowycz (changes requested):
hi! make sure to review the [Checklist](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md), then apply some fixes to your work

src/index.html
```diff
@@ -1,16 +1,59 @@
...
+          name="search--small"
+          placeholder="Try &quot;Los Angeles&quot;"
+          type="text"
+        >
+      </label>
+    </form>
+
   </body>
+
 </html>
```
jstmpelowycz, line 59: remove extra whitespaces

src/style.css
```diff
@@ -0,0 +1,95 @@
+@font-face {
+  font-family: "Avenir-Book";
+  src: url("fonts/Avenir-Book.ttf");
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url("fonts/Avenir-Heavy.ttf");
+}
```
jstmpelowycz, line 9: there is no need to give distinct names to these font-faces, name should be the same so that it can be used the following way:
```
.class1 { // uses Avenir-Book
  font-family: Avenir, ...;
  font-weight: 300;
}

.class2 { // uses Avenir-Heavy
  font-family: Avenir, ...;
  font-weight: 900;
}
```

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+
+.wrapper {
+  position: relative;
```
jstmpelowycz, line 12: no need for relative positioning

src/style.css
jstmpelowycz: also, you'd rather put `font-style: normal` here to avoid putting it several times down below

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+}
+
+.search::placeholder {
+  color: #3d4e61;
+  font-style: normal;
+  position: absolute;
+  left: 0%;
+  right: 18.35%;
+  top: 0%;
+  bottom: -15.79%;
```
jstmpelowycz, line 43: absolute positioning is redundant here

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+  font-size: 16px;
+  line-height: 22px;
+  width: 129px;
+  height: 22px;
```
jstmpelowycz, line 49: setting width and height here is redundant

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+.search:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+  font-family: "Avenir", sans-serif;
+  font-style: normal;
```
jstmpelowycz, line 62: do not duplicate styles, you've already put them up above

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  font-weight: 900;
+  color: #3d4e61;
+  font-size: 16px;
+  line-height: 22px;
+  padding: 0 677px 0 62px;
```
jstmpelowycz, line 67: you don't need this `677px`

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  height: 11px;
+}
+
+.search--small::placeholder {
+  font-family: "Avenir-Book", sans-serif;
+  font-weight: 300;
+  font-size: 14px;
+  line-height: 19px;
+  width: 167.29px;
+  height: 20px;
```
jstmpelowycz, line 87: setting width and height here is redundant here

jstmpelowycz (changes requested):
almost done! let's resolve the issues

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+
+.search-icon2 {
+  top: 15px;
+  left: 13px;
+  width: 11px;
+  height: 11px;
+}
+
+.search--small::placeholder {
+  font-family: "Avenir-Book", sans-serif;
```
jstmpelowycz, line 80: you don't have Avenir-Book font, replace it with Avenir

src/style.css
jstmpelowycz: fix it, please

src/style.css
jstmpelowycz: fix it, please

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+}
+
+.search:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+  font-family: "Avenir", sans-serif;
```
jstmpelowycz, line 60: put declaring this rule under `.search` selector
by doing so, you won't have to duplicate it several times

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+  font-weight: 900;
+  color: #3d4e61;
```
jstmpelowycz, line 62: you've already put this color rule up above

jstmpelowycz (changes requested):
you've got a few issues:
- when clicking on search, it jumps right, and after unfocusing – jumps back, it's not OK
- font-weight does not get bold when search input gets focused

you can observe the example of this work and how it should look like [here](https://jstmpelowycz.github.io/layout_search-bar-airbnb/)

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  padding-left: 33px;
+}
+
+.search--small::placeholder {
+  width: 167px;
+  height: 20px;
+}
+
+.search--small:focus {
+  padding-left: 33px;
```
amitpatiljc, line 75: you already have ` padding-left` as default style in `.search--small ` class you don need it here

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+
+.search:hover {
+  color: #3d4e61;
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  font-family: "Avenir-Heavy", sans-serif;
+  color: #3d4e61;
```
amitpatiljc, line 50: add `color` property to default class `.search` so that you dont have to add on `.search:focus ` and `.search:hover `

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  font-weight: 300px;
+  font-size: 16px;
+  line-height: 22px;
+  box-sizing: border-box;
+  height: 70px;
+  padding-left: 62px;
+}
+
+.search:hover {
+  color: #3d4e61;
```
amitpatiljc, line 43: add color property to default class `.search `so that you dont have to add on `.search:focus` and `.search:hover`

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+.search--small {
+  height: 42px;
+  width: 100%;
+  font-size: 14px;
+  padding-left: 33px;
+}
+
+.search--small::placeholder {
+  width: 167px;
+  height: 20px;
```
amitpatiljc, line 71: do not use hard coded `height` and `width`  instead use `font-size` and `line-height`

amitpatiljc (changes requested):

src/style.css
Alessandranitti15: Hello Amit, if I delete padding-left in :focus, then the course jumps forward.

PR #2118, opened 2022-09-29
first commit

44shaggy:
- [DEMO LINK](https://44shaggy.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://44shaggy.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,104 @@
+
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "AvenirHeavy";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 900;
+}
+
+.searchbar {
+  position: absolute;
+  box-sizing: border-box;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  width: 100%;
+  left: 0%;
+  right: 0%;
+  top: 0%;
+  bottom: 0%;
+}
+
+.searchbar.big {
+  height: 70px;
+  top: 20px;
+}
+
+.searchbar.small {
+  height: 42px;
+  top: 110px;
+}
+
+.searchbar:hover {
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.searchbar:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+}
+
+.text {
+  position: absolute;
+  border-width: 0;
+  font-family: "Avenir", sans-serif;
+  font-weight: 900;
+  color: #3d4e61;
+  width: 158px;
+}
+
+.text.big {
+  font-size: 16px;
+  line-height: 22px;
+  left: 62px;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,104 @@
+
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "AvenirHeavy";
```
jstmpelowycz, line 10: this font should have the same name – _Avenir_, just change its font-weight so that it can be used the following way
```
.class1 {
  font-family: Avenir, ...;
  font-weight: 300;
}

.class2 {
  font-family: Avenir, ...;
  font-weight: 900;
}
```

src/style.css
```diff
@@ -0,0 +1,104 @@
...
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 900;
+}
+
+.searchbar {
+  position: absolute;
```
jstmpelowycz, line 17: that is the cause preventing your app from passing tests
do not use absolute positioning, it literally makes your elements stuck in one place, as a result – it is not possible to dress them up with paddings and margins
what you can do – for example, you can try applying horizontal paddings to your forms so that they do not adhere to the left/right sides

src/style.css
```diff
@@ -0,0 +1,104 @@
...
+  border-radius: 3px;
+}
+
+.text {
+  position: absolute;
+  border-width: 0;
+  font-family: "Avenir", sans-serif;
+  font-weight: 900;
+  color: #3d4e61;
+  width: 158px;
```
jstmpelowycz, line 56: you don't need it

src/style.css
```diff
@@ -0,0 +1,104 @@
...
+}
+
+.text.big {
+  font-size: 16px;
+  line-height: 22px;
+  left: 62px;
+  top: 35.71%;
+  bottom: 37.14%;
```
jstmpelowycz, line 64: for sure, that is inappropriate to use such values, with decimals and hundreds

src/style.css
```diff
@@ -0,0 +1,104 @@
...
+
+.icon {
+  position: absolute;
+}
+
+.icon.big {
+  width: 19px;
+  height: 19px;
+  left: 26px;
+  top: 25px;
```
jstmpelowycz, line 82: after removing absolute positioning, don't forget to remove `left`, `right`, `top`, `bottom`

jstmpelowycz (changes requested):
let's apply some changes

44shaggy:
I might've followed the task instruction "follow styles from the mock" too literally. That's where i got absolute positioning and strange decimal values in the first place.

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+.searchbar.small:focus {
+  background: url(./images/Search.svg) no-repeat 12px 14px / 11px, linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
+
+.searchbar::placeholder {
+  font-family: "Avenir", sans-serif;
+  color: #3d4e61;
+}
+
+.searchbar:focus::placeholder {
```
amitpatiljc, line 61: on focus placeholder do not change at all, only when you write in it, that text should be bold 
but in your case when focus on a particular input box that placeholder's font becomes bold

amitpatiljc (changes requested):

src/style.css
44shaggy: easy peasy ;)

amitpatiljc (approved):
Good Job!!

PR #2129, opened 2022-10-03
Develop

nadiia-kurovska:
- [DEMO LINK](https://nadiia-kurovska.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://nadiia-kurovska.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,94 @@
+@font-face {
+  font-family: "Avenir-Book", sans-serif;
+  src: url("fonts/Avenir-Book.ttf") format("truetype");
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir-Heavy", sans-serif;
+  src: url("fonts/Avenir-Heavy.ttf") format("truetype");
+  font-weight: 300;
+}
+
+* {
+  box-sizing: border-box;
+}
+  
+.search-form {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,94 @@
...
+  left: 26px;
+  top: calc(50% - 11px);
+  text-align: center;
+  line-height: 22px;
+  width: 19px;
+  height: 19px;
+}
+  
+.search-form-small::after {
+  position: absolute;
```
qoa11a, line 77: You have this property in .search-form::after so you don't need to repeat it here.

qoa11a (changes requested):
image
You have failed tests. From what I can see the problem is in fonts/margins. You should fix it

src/style.css
nadiia-kurovska: I have fixed the issue with on line 77. Thank you. 
As for the failed tests, I can see that most likely the issue is with font, however, all the attempts i've made gave no result. 
Is there anything you could recommend?

nadiia-kurovska: fixed what i could for now. now at least it passes all the tests.

qoa11a (approved):
Good job!

PR #2146, opened 2022-10-07
task implemented

Alessandranitti15:
https://Alessandranitti15.github.io/layout_search-bar-airbnb/
https://Alessandranitti15.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,71 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 900;
+}
+
+.wrapper {
+  position: relative;
+  display: block;
+  margin-top: 20px;
+  min-width: 100px;
+}
+
+.icon {
+  position: absolute;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,71 @@
...
+  height: 42px;
+  width: 100%;
+  font-size: 14px;
+  padding-left: 33px;
+  line-height: 20px;
+}
+
+.search--small:focus {
+  padding-left: 33px;
+}
```
amitpatiljc, line 71: in focus state, as per figma design placeholder should not change font-weight but what ever form input we write that should be bold 
but in your case when we click on input field place holder also becomes bold 
so please consider the last point in the checklist
`Remember that placeholder has its own set of styles available using ::placeholder pseudo-element.`

amitpatiljc (changes requested):

src/style.css
```diff
@@ -0,0 +1,76 @@
...
```
amitpatiljc, line 76: for class `.search--small` you already have `padding-left: 33px;` focus state will inherit from default so IMO this code is redundant

amitpatiljc (changes requested):
Just one change and done!

src/style.css
Alessandranitti15: oh sorry, I have forgot. now is fixed, thanks

jstmpelowycz (approved):
GJ 🔥

PR #2151, opened 2022-10-09
Develop

domkab:
- [DEMO LINK] (https://domkab.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK] (https://domkab.github.io/layout_search-bar-airbnb/report/html_report/)


src/index.html
```diff
@@ -1,16 +1,59 @@
...
+      >
+        <input
+          data-qa="keypress"
+          class="search search_big"
+          type="text"
+          id="search_big"
+          placeholder="Try &quot;Los Angeles&quot"
+        >
+      </form>
+    </label>
```
amitpatiljc, line 39: the HTML structure  is wrong, there should be` label` for `input` not for `form`
so it should be form - label - input

src/style.css
```diff
@@ -0,0 +1,63 @@
...
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: bold;
+}
+
+* {
+  margin-top: 0;
```
amitpatiljc, line 14: there is no default margin for form so this line of code is redundant

amitpatiljc (commented):
Implement HTML structure as mentioned and only add and deploy files inside src folder
do no add json files

amitpatiljc (changes requested):
Implement HTML structure as mentioned and only add and deploy files inside src folder
do no add json files

src/style.css
domkab: @amitpatiljc why? I delete it and the top margin moves up which is making me fail the only test case I pass

amitpatiljc (changes requested):

src/index.html
```diff
@@ -1,16 +1,60 @@
...
+          type="text"
+          id="search_small"
+          placeholder="Try &quot;Los Angeles&quot"
+        >
+      </label>
+
+    </form>
   </body>
+
 </html>
```
jstmpelowycz, line 60: some of the spaces are just redundant
```suggestion
    <form
      class="form_box"
      id="form"
      data-qa="small"
    >
      <label for="search_small">
        <input
          class="search search_small"
          type="text"
          id="search_small"
          placeholder="Try &quot;Los Angeles&quot"
        >
      </label>
    </form>
  </body>
</html>
```

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.search:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.search:focus {
+  font-weight: bold;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
jstmpelowycz, line 43: try to use `background-color` property instead of `background` – it causes either icon to disappear if user focuses on it

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+
```
jstmpelowycz, line 44: don't forget to remove redundant spaces 🙂

jstmpelowycz (changes requested):
almost done, @domkab, let's apply a couple of fixes

jstmpelowycz (approved):
nice!

PR #2155, opened 2022-10-11
Added search bar

nazarkharkevych:
https://nazarkharkevych.github.io/layout_search-bar-airbnb/
https://nazarkharkevych.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,70 @@
+:root {
+  --background-image: url(./images/Search.svg) no-repeat 26px 25px;
+}
+
+body {
+  margin-top: 0;
+  margin-bottom: 0;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 900;
+}
+
+.form {
+  margin-top: 20px;
+}
+
+.input {
+  font-family: Avenir, sans-serif;
+  font-weight: 900;
+  font-size: 16px;
+  line-height: 22px;
+  width: 100%;
+  height: 70px;
+  /* margin-top: 20px; */
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,70 @@
+:root {
+  --background-image: url(./images/Search.svg) no-repeat 26px 25px;
+}
+
+body {
+  margin-top: 0;
+  margin-bottom: 0;
```
SanyaBratashchuk, line 7:
```suggestion
  margin: 0;
```

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  font-style: normal;
+  font-weight: 900;
+}
+
+.form {
+  margin-top: 20px;
+}
+
+.input {
+  font-family: Avenir, sans-serif;
```
SanyaBratashchuk, line 29: move it to the `html` tag selector and let input inherit it

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  font-weight: 900;
+  font-size: 16px;
+  line-height: 22px;
+  width: 100%;
+  height: 70px;
+  /* margin-top: 20px; */
```
SanyaBratashchuk, line 35: do not leave commetns

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.input:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.input:focus {
+  border-radius: 3px;
+  background: var(--background-image),
+    linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
SanyaBratashchuk, line 64:
```suggestion
  linear-gradient(180deg, #fff 0%, #f6f6f7 100%);

```

SanyaBratashchuk (changes requested):
Remove this outline
image

src/style.css
nazarkharkevych: Lint expects the indentation of 4 spaces on this line and did not accept two

src/style.css
nazarkharkevych: margin: 0; alone fails the tests

SanyaBratashchuk (approved):

PR #2189, opened 2022-10-19
changed index.html and styles.html. Added full code

ZupaFly:
- [DEMO LINK](https://ZupaFly.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ZupaFly.github.io/layout_search-bar-airbnb/report/html_report/)


jstmpelowycz (changes requested):
nice visual, as expected! 
about code – let's create a common class for common elements and ensure there are no code duplications

src/index.html
```diff
@@ -1,16 +1,39 @@
...
+        method="get"
+      >
+        <input 
+          type="search"
+          placeholder="Try &#34Los Angeles&quot"
+          data-qa="big"
+          class="bigSearch"
+        >
+      </form>
+    </label>
```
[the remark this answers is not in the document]
ZupaFly: I am changing my code according to your comments 1 form 2 inputs. But before i did 2 forms because of the task: Each search bar be inside its own `form` element. (from Read me file). I guess it's because for positioning 2 elements i am using margin top/bottom. And because of <label> tag. Margin part is also active while i am pointing on it. So now 1 element have active zone of it's own height + margin and another one is just

src/index.html
```diff
@@ -1,16 +1,39 @@
...
+    <label class="bigBar">
+      <form 
+        action="server.name" 
+        method="get"
+      >
+        <input 
+          type="search"
+          placeholder="Try &#34Los Angeles&quot"
+          data-qa="big"
+          class="bigSearch"
```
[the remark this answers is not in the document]
ZupaFly: Done according to the comments

src/index.html
```diff
@@ -1,16 +1,31 @@
...
+          class="search search__big"
+        >
+        <input
+          type="search"
+          placeholder="Try &#34Los Angeles&quot"
+          data-qa="small"
+          class="search search__small"
+        >
+      </label>
+    </form>
```
jstmpelowycz, line 29: in this case with search bars, there should be two forms (+ don't forget about `data-qa` attributes to make sure it is possible to auto-test you solution)
```
form --big
	label
		input

form --small
	label
		input

File: src/index.html (hunk 1 above)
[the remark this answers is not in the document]
Comment (jstmpelowycz) [reply src/index.html]: no, there still should be two forms, you'd just mixed up with the order of tags
it is just an example, but the order is
```
<form>
  <label>
    <input/>
  </label>
  <label>
    <input/>
  </label>
  <label>
    <input/>
  </label>
</form>

<form>
  <label>
    <input/>
  </label>
</form>
```

**Review by jstmpelowycz — CHANGES_REQUESTED** [review]
great job! last hint – and you're done!

File: src/style.css [hunk 4]
```diff
@@ -0,0 +1,61 @@
...
+
+@font-face {
+  font-family: Avenir-Book;
+  src: url(fonts\Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir-Heavy;
+  src: url(fonts\Avenir-Heavy.ttf);
+}
```
jstmpelowycz, line 14: here is a better approach: font-family is the same, we just indicate font-weight (so that we know what font to use by indicating its font-weight)
```suggestion
@font-face {
  font-family: Avenir;
  src: url(fonts\Avenir-Book.ttf);
  font-weight: normal;
}

@font-face {
  font-family: Avenir;
  src: url(fonts\Avenir-Heavy.ttf);
  font-weight: bold;
}
```

jstmpelowycz (approved):
Great!

src/style.css
ZupaFly: I got it, will implement it to my code. Should i push it once again after implementing changes?

jstmpelowycz: yes, let's commit changes, push them, and then re-run deploy

ZupaFly: Done

PR #2207, opened 2022-10-21
Done

Nazar-Brunarskyi:
- [DEMO LINK](https://Nazar-Brunarskyi.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Nazar-Brunarskyi.github.io/layout_search-bar-airbnb/report/html_report/)


src/index.html
```diff
@@ -10,7 +10,46 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+  
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="/"
```
valerii-derkach, line 16:
```suggestion
      action="#"
```
try to change and find out what is the difference

valerii-derkach (changes requested):
checklist point 2

src/index.html
Nazar-Brunarskyi: I have found something like this on StackOverflow

action="#" is the same as link to href="#". It doesn't go anywhere but at the top of that page. The action="/" is the same as link to href="/". The single slash / is root web URL

valerii-derkach: and when you submit the form when action='/', we have error 404, and when # - just reload page
also u can use the link from previous task

valerii-derkach (approved):

PR #2210, opened 2022-10-22
first commit

kenvastolord:
- [DEMO LINK](https://kenvastolord.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://kenvastolord.github.io/layout_search-bar-airbnb/report/html_report/)


meforoma (approved):
good structure,
passable styling, but you could try:
instead of setting the height explicitly,
- you can try to set the correct padding, font size, and line-height => the height will then be calculated automatically

PR #2246, opened 2022-10-25
Create HTML page with two search bars

NadiiaKoch:
https://NadiiaKoch.github.io/layout_search-bar-airbnb/
https://NadiiaKoch.github.io/layout_search-bar-airbnb/report/html_report/


valerii-derkach (approved):
Good job!

PR #2253, opened 2022-10-26
Develop

Lun4trick:
https://Lun4trick.github.io/layout_search-bar-airbnb/
https://Lun4trick.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -57,7 +57,7 @@ body {
 
 }
 
-.searchBar-small {
+.searchBar--small {
   height: 42px;
   background-position-x: 13px;
   background-position-y: 15px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  margin-bottom: 20px;
+  color: rgba(61, 78, 97, 1);
+}
+
+.searchBar:hover {
+  left: 0%;
+  right: 0%;
+  top: 0%;
+  bottom: 0%;
+  border: 1px solid #e1e7ed;
```
naziamoff, line 42: looks like you already have those rules inside `.searchBar`, do you really need them here too?

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.searchBar:focus {
```
naziamoff, line 47: same here, check for the repeated css-rules

naziamoff (approved):
Good job overall)
There's one thing we can and should improve

PR #2283, opened 2022-10-31
searchbar added

UmizDemud:


src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  margin: 0;
+  padding: 0;
+  border: none;
+  background: none;
+}
+
+.big .search-button {
+  top: 25px;
+  left: 26px;
+}
```
naziamoff, line 60: nice work with the selectors here 👍 
still there's a better way of doing it)

you can make one selector for the searchbar and add the repeated rules there
```
.searchBar {
** all repeated styling + styling for big searchbar goes here **
}

.searchBar--small {
** here goes styling that differs **
}
```

naziamoff (approved):
Good job! There are some comments though

PR #2291, opened 2022-11-01
solved the task

MeryamWazeen:
- [DEMO LINK](https://meryamwazeen.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://meryamwazeen.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,51 @@
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
```
jstmpelowycz, line 5: there should be 2 font-faces with the same name, the only way to select either is by using font-weight, in this case: if `normal` – we use Avenir-Book, if `bold` – we use Avenir-Heavy
```
@font-face {
  font-family: Avenir;
  src: url(fonts/Avenir-Book.ttf);
  font-weight: normal;
}

@font-face {
  font-family: Avenir;
  src: url(fonts/Avenir-Heavy.ttf);
  font-weight: bold;
}
```

jstmpelowycz (changes requested):
Great job! There are just a few issues:
- font-weight does not get bold on focus
- the cause of failed tests – the margins of `body`, just remove them
Screenshot 2022-11-01 at 15 33 50

src/style.css
```diff
@@ -0,0 +1,62 @@
...
+  background-position: 12px 12px;
+}
+
+.search:hover {
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  font-family: Avenir-Book, sans-serif;
```
jstmpelowycz, line 59:
```suggestion
  font-weight: bold;
```

src/style.css
```diff
@@ -0,0 +1,62 @@
...
+
+}
+
+.search {
+  width: 100%;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  font-weight: 300px;
+  color: #3d4e61;
+  font-family: Avenir-Book, sans-serif;
```
jstmpelowycz, line 24: there is no such font-family, you've described only Avenir + its two font-weights, so where Avenir-Book is supposed to come from? right – it won't happen, so that you have to use Avenir with normal font-weight

jstmpelowycz (changes requested):
almost done

src/style.css
jstmpelowycz: I don't see that you've committed your changes, probably – you've forgotten to run `git commit` and `git push` commands

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  font-weight: bold;
+  border-radius: 3px;
+  border: 1px solid rgba(255, 231, 237, 1);
+}
+
+.search:default {
```
jstmpelowycz, line 65: what is `:default` for here? do you need it?

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  position: absolute;
+  left: 0%;
+  right: 18.35%;
+  top: 0%;
+  bottom: -15.79%;
```
jstmpelowycz, line 70: suggestion: do not use such values in the future, we call them ✨ Magic Numbers ✨ – none can understand where they come from, moreover, if there are decimals, hundredths

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  font-family: Avenir-Book, sans-serif;
```
jstmpelowycz, line 71: there is still no such font Avenir-Book available, use Avenir

jstmpelowycz (changes requested):

jstmpelowycz (commented):
great! you've forgotten to re-deploy, but OK, just don't forget to do it next time 🙂

jstmpelowycz (approved):

PR #2293, opened 2022-11-01
added search bars and corresponding styles, updated readme.md

Rijo-J:
- [DEMO LINK](https://Rijo-J.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Rijo-J.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,103 @@
+* {
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 600;
+}
+
+@font-face {
+  font-family: "Avenir";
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,103 @@
...
+  margin: 20px auto 0;
+}
+
+.big {
+  height: 70px;
+}
+
+.small {
+  height: 42px;
+}
```
naziamoff, line 36: you can make it like that and we will have more reusable stuff :)
And one more point - we use modifiers with `--`. Like `form--big` `form--small`  and so on. You will learn about it later in the topic BEM
```suggestion
.form {
  display: block;
  height: 70px;
  justify-content: center;
  margin: 20px auto 0;
}

.form--small {
  height: 42px;
}
```

naziamoff (approved):
Well done 🔥 
There are some comments

PR #2296, opened 2022-11-01
search bar was created

skorykdariia:
- [DEMO LINK](https://skorykdariia.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://skorykdariia.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,5 +1,6 @@
 <!doctype html>
 <html lang="en">
+
   <head>
     <meta charset="UTF-8">
     <meta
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -10,7 +11,48 @@
...
+      data-qa="big"
+      class="bigForm form"
+      action="http://localhost:8080/"
+      method="post"
+    >
+      <img
+        class="bigIcon"
+        src="./images/Search.svg"
+        alt="icon"
+      >
```
Ihor-Karpyn, line 26: better to use background image

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  font-weight: 900;
+}
+
+html {
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+}
+
+.form {
+  position: relative;
```
Ihor-Karpyn, line 20: Don't add position on the form add some wrapper for input. If u add another input inside the form, the icon will break

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+}
+
+.bigForm {
+  margin: 20px auto;
+}
+
+.firstBar {
```
Ihor-Karpyn, line 27: let's use big and small. The class name shouldn't depend on the layout.

Ihor-Karpyn (changes requested):
Something wrong with font, compare with markup

image

Ihor-Karpyn (approved):

PR #2302, opened 2022-11-01
commit number 1

lukacsja:
- [DEMO LINK](https://lukacsja.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://lukacsja.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,117 @@
...
+  font-size: 16px;
+  line-height: 22px;
+  color: #3d4e61;
+  width: 100%;
+  background-color: #fff;
+  box-shadow: 0 1px 8px rgb(61 78 97 / 10%);
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  font-family: Avenir, sans-serif;
+}
```
naziamoff, line 36: there's a bunch of css-rules that are repeated in input__big and input__small.
Better to create separate rule for just `.input` where you add everything reusable.

lukacsja: I had to git clone this again (made backup) then tried to push it back but could not because already exist (i guess?).
May I have some assist in this matter please?

SalehBUD: If you need some help with this issue, maybe I can help. Just message me in slack 👋🏻

PR #2304, opened 2022-11-01
First try

4ebupel:
- [DEMO LINK](https://4ebupel.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://4ebupel.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,85 @@
...
+  font-weight: 800;
+}
+
+html {
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+}
+
+.form {
+  position: relative;
```
Ihor-Karpyn, line 21: Do not add relative to the form, add some wrapper for input, if u add another input inside the form, the icon will break

src/style.css
```diff
@@ -0,0 +1,85 @@
...
+  margin: 20px auto;
+}
+
+.first__input {
```
Ihor-Karpyn, line 25: The class shouldn’t depend on HTML structure, if I swap input places should I rename this class? Let's use big and small

Ihor-Karpyn (approved):

PR #2316, opened 2022-11-02
Search bar for Airbnd

dahimohamed:
- [DEMO LINK](https://dahimohamed.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://dahimohamed.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -71,7 +71,6 @@ body {
 
 .form__input:focus {
   background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
-  filter: drop-shadow(0 4px 4px rgba(0, 0, 0, 0.25));
   border-radius: 3px;
   font-weight: 800;
 }


src/index.html
```diff
@@ -11,6 +11,51 @@
...
   <body>
-    <h1>Search bar airbnb</h1>
+    <!-- <h1>Search bar airbnb</h1> -->
+    <form
+      class="form"
+      action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+      method="post"
+    >
+
+      <img
```
naziamoff, line 21: you should not add empty lines if elements are on the different levels of nesting
```suggestion
      method="post"
    >
      <img
```

src/index.html
```diff
@@ -11,6 +11,51 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <!-- <h1>Search bar airbnb</h1> -->
```
naziamoff, line 14: always make sure that you removed all the unnecessary comments

naziamoff (changes requested):
Good job! There's but one thing that is critical to improve
- it looks wrong if focused
Screenshot 2022-11-03 at 17 45 26

Also, there's a bunch of small comments that would be nice to have to fix

src/index.html
Mohameddahi: Thank you for your feedback.
now , it is updated. Please check it !

Mohameddahi: @naziamoff

naziamoff (approved):
Well done 🔥

PR #2318, opened 2022-11-02
Finished search bar

JzSol:
- [DEMO LINK](https://JzSol.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://JzSol.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,8 +12,6 @@
   </head>
 
   <body>
-    <!-- <h1>Search bar airbnb</h1> -->
-
     <form
       action="https://mate-academy-form-lesson.herokuapp.com/create-application"
       method="get"
--- a/src/index.html
+++ b/src/index.html
@@ -1,4 +1,4 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -10,7 +10,40 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <!-- <h1>Search bar airbnb</h1> -->
+
```
naziamoff, line 16: - unnecessary comments should be removed
- there shouldn't be empty lines between parent and child elements

src/style.css
```diff
@@ -0,0 +1,115 @@
...
+  padding-left: 62px;
+  font-size: 16px;
+  line-height: 22px;
+}
+
+.search-bar--small {
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 19px;
+}
```
naziamoff, line 95: you can make one selector for the searchbar and add the repeated rules and the rules from the big searchbar there
```
.searchBar {
** all repeated styling + styling for big searchbar goes here **
}

.searchBar--small {
** here goes styling that differs **
}
```

naziamoff (changes requested):
Looks good to me 👍 
Fix my comments and add the demo and the test report links to the PR description

src/style.css
```diff
@@ -0,0 +1,115 @@
...
+
+.search-label--small::after {
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
+
+
+
```
naziamoff, line 70: too much empty lines here, you need only one

src/style.css
```diff
@@ -0,0 +1,115 @@
...
+  width: 100%;
+}
+
+.search-form--big {
+  height: 70px;
+}
+
+.search-form--small {
+  height: 42px;
+}
```
naziamoff, line 37: why don't we add the height of the big form right to the `.search-form` selector? In that case we won't need two additional selectors

src/style.css
```diff
@@ -0,0 +1,115 @@
...
+  top: 25px;
+  left: 26px;
+}
+
+.search-label--small::after {
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
```
naziamoff, line 67: same here

src/style.css
```diff
@@ -0,0 +1,115 @@
...
+
+
+
+.search-bar:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.searchBar {
+width: 100%;
```
naziamoff, line 76: don't forget to add 2 indents

src/style.css
```diff
@@ -0,0 +1,115 @@
...
+  line-height: 22px;
+}
+
+.search-bar--big::placeholder {
+  font-size: 16px;
+}
+
+.search-bar--small::placeholder {
+  font-size: 14px;
+}
```
naziamoff, line 109: and here

naziamoff (changes requested):

naziamoff (approved):
Nice! 🔥

PR #2324, opened 2022-11-02
save-01

1veee:


alexandra-protyanova (changes requested):
Add demo and test report links to your pull request

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+  
+* {
+  box-sizing: border-box;
+}
+  
+body {
+  margin-top: 20px;
+}
+  
+.seachBar {
```
ivanchuk-yurii, line 21: use kebab-case, not camelCase for class naming, should be `search-bar`

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+  
+.seachBar_input {
+  width: 100%;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  
+  color: #3d4e61;
+  
+  font-family: Avenir, sans-serif;
```
ivanchuk-yurii, line 51: add this to `html` and inherit in input

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+.seachBar_input::placeholder {
+  color: #3d4e61;
+  font-weight: 300;
+}
+  
+.seachBar_input:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+  
+.seachBar_input:focus {
```
ivanchuk-yurii, line 64: fix your `:focus` styles, they should be as on mockup

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+.seachBar:before {
+  position: absolute;
+  
+  display: block;
+  content: "";
+  
+  background-image: url("images/Search.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: contain;
```
ivanchuk-yurii, line 79: combine into one `background` property

ivanchuk-yurii (changes requested):
links should be here
image

PR #2326, opened 2022-11-02
add search bars

m-hrytsenko:
- [DEMO LINK](https://m-hrytsenko.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://m-hrytsenko.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,131 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,131 @@
...
+}
+
+.formSmall {
+  height: 42px;
+  line-height: 19px;
+}
+
+.search {
+  position: relative;
+  box-sizing: border-box;
```
polosanya, line 39: It is a good tone to set this property for all elements(*) and don't change without extra need

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comments)

src/style.css
m-hrytsenko: Thanks for your advice!
I'll keep that in mind next time)

PR #2390, opened 2022-11-13
Added 2 versions of the search bar

bojkovladislav:


src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
```
SanyaBratashchuk, line 21: image

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+
+.search-image--small {
+  height: var(--small-image-height);
+  top: calc((var(--rectangle-small-height) - var(--small-image-height)) / 2);
+  left: 13px;
+}
+
+.input {
+  font-family: "Avenir-Book", Arial, Helvetica, sans-serif;
```
SanyaBratashchuk, line 49: move it to the html tag selector and let inputs inherit it 
```suggestion
  font-family: inherit;
```

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  color: #3d4e61;
+  outline: 0;
+}
+
+.input:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.input:focus {
+  font-family: "Avenir-Heavy", Arial, Helvetica, sans-serif;
```
SanyaBratashchuk, line 64:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.input-small {
+  height: var(--rectangle-small-height);
+  padding-left: 33px;
+  font-size: 14px;
+}
+
+.input::placeholder {
+  font-family: "Avenir-Book", Arial, Helvetica, sans-serif;
```
SanyaBratashchuk, line 82:
```suggestion
```

SanyaBratashchuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+  margin: 20px 0;
+}
+
+.search-image {
+  position: absolute;
+}
+
+.search-image--big {
+  height: var(--big-image-height);
+  top: calc((var(--rectangle-big-height) - var(--big-image-height)) / 2);
```
ivanchuk-yurii, line 40: it's better to use `%`

ivanchuk-yurii (approved):

src/style.css
bojkovladislav: Code_lXcFrASorJ

bojkovladislav: What am I doing wrong?

PR #2395, opened 2022-11-14
made search bar

sv-dubov:
- [DEMO LINK](https://sv-dubov.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://sv-dubov.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,7 +15,8 @@
       action="http://localhost:8080"
       class="form"
       data-qa="big"
-      method="get">
+      method="get"
+    >
       <label>
         <img
           src="./images/Search.svg"
@@ -24,7 +25,7 @@
         >
         <input
           type="text"
-          class="search__bar big"
+          class="search__bar search__bar--big"
           name="searchBig"
           data-qa="keypress"
           placeholder="Try &quot;Los Angeles&quot;"
@@ -35,7 +36,8 @@
       action="http://localhost:8080"
       class="form"
       data-qa="small"
-      method="get">
+      method="get"
+    >
       <label>
         <img
           src="./images/Search.svg"
@@ -44,7 +46,7 @@
         >
         <input
           type="text"
-          class="search__bar small"
+          class="search__bar search__bar--small"
           name="searchSmall"
           data-qa="keypress"
           placeholder="Try &quot;Los Angeles&quot;"
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,45 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,45 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="http://localhost:8080"
+      class="form"
+      data-qa="big"
+      method="get">
```
ArturErmolenko, line 18:
```suggestion
    <form
      action="http://localhost:8080"
      class="form"
      data-qa="big"
      method="get"
    >
```

src/index.html
```diff
@@ -11,6 +11,45 @@
...
+          data-qa="keypress"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
+    <form
+      action="http://localhost:8080"
+      class="form"
+      data-qa="small"
+      method="get">
```
ArturErmolenko, line 38:
```suggestion
    <form
      action="http://localhost:8080"
      class="form"
      data-qa="small"
      method="get"
    >
```

src/index.html
```diff
@@ -11,6 +11,45 @@
...
+      <label>
+        <img
+          src="./images/Search.svg"
+          alt="Search icon"
+          class="icon icon-big"
+        >
+        <input
+          type="text"
+          class="search__bar big"
```
ArturErmolenko, line 27:
```suggestion
          class="search__bar"
```
checklist point 2

ArturErmolenko (changes requested):

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+  font-style: normal;
+  font-weight: 300;
+  color: #3d4e61;
+}
+
+.search__bar--big {
+  height: 70px;
+  font-size: 16px;
+  padding-left: 62px;
+}
```
ArturErmolenko, line 51: checklist point 2
if you have two or more similar elements with portions of similar styles with different values - use one of the elements as the basic case, and override necessary styles for other cases. Explanation: The point is not in the names of the classes, the point is: when there are several similar elements, ex., 2 inputs, for one we can give a class input, for example, and for the second - input input--small. We write all the styles for .input, but for .input-small we write only those styles that differ in design, and we need this second input to look a little different. Element with class .input without extra classes should also look like a full-fledged styled element.

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #2403, opened 2022-11-15
Develop

andrewchornobai:


SanyaBratashchuk (changes requested):
input should be fullwidth 
image

PR #2405, opened 2022-11-16
Search-bar Rybak Dmytro

fisherok3452:
- [DEMO LINK](https://fisherok3452.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://fisherok3452.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+.bar {
+  position: relative;
+  margin-top: 20px;
```
Viktor-Kost, line 18: If you are going to give a distance between two forms, you should give a margin to a parent element (form), not to a child one

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+}
+
+body {
+  font-family: Avenir, Helvetica, sans-serif;
+}
+
+.firstBarInput {
```
Viktor-Kost, line 25: Try to follow DRY principe (Don't repeat yourself). So here u can collect such rules as  border, box-shadow, border-radius, etc
And use them like this 
```
.firstBarInput,
.secondBarInput {
....your common rules
}
```

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  text-indent: 33px;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  padding: 0;
+  color: #3d4e61;
+  font-family: Avenir-Book, Helvetica, sans-serif;
+}
+
+input::placeholder {
```
Viktor-Kost, line 51: It would better use class selector instead of tag one

Viktor-Kost (changes requested):
Nice!) 
Please, make some improvements

Virass (approved):

PR #2407, opened 2022-11-16
Search bars

khrystyna-prystai:


src/style.css
```diff
@@ -0,0 +1,83 @@
...
+
+@font-face {
+  font-family: "AvenirHeavy";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+}
+
+* {
+  box-sizing: border-box;
+  font-weight: 300;
```
loralevitska, line 15: move to first font-face

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.largebar {
+  height: 70px;
+  padding-left: 62px;
+
+}
+
+.largebar::placeholder {
+  font-family: "AvenirBook", sans-serif;
```
loralevitska, line 41: move to html selector and remove repeats in other classes, instead this inherit it

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+  font-weight: 300;
+}
+
+.form {
+  margin-top: 20px;
+}
+
+.form1 {
+  margin-top: 20px;
+}
```
loralevitska, line 24: the styles are identical, so for another form you can apply this class `form` too and remove `form1`

src/style.css
```diff
@@ -0,0 +1,83 @@
+@font-face {
+  font-family: "AvenirBook";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
```
loralevitska, line 4: this is the default value, so you don't need to set it again

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+@font-face {
+  font-family: "AvenirHeavy";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
```
loralevitska, line 10: same here. Remove it

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.largebar::placeholder {
+  font-family: "AvenirBook", sans-serif;
+  font-size: 16px;
+  line-height: 22px;
+  color: #3d4e61;
+}
+
+.image1 {
```
loralevitska, line 47: if you don't read the task, it's not clear from the code what `image1` `image2` is. Use more understandable variables, for example, `largeIcon` etc...

Viktor-Kost (approved):
Fine!)

PR #2408, opened 2022-11-16
Created big and small search bars

TkachYuliia:
- [DEMO LINK](https://TkachYuliia.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://TkachYuliia.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,116 @@
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
+
+html {
+  font-family: Avenir, sans-serif;
+  font-style: normal;
+}
+
+.search {
+  height: 70px;
+  margin: 20px 0;
+}
+
+.searchSmall {
+  height: 42px;
+}
+
+.img {
+  width: 19px;
+}
+
+.imgSmall {
+  width: 11px;
+}
+
+.input {
+  box-sizing: border-box;
+  width: 100%;
+  outline: none;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  font-style: normal;
+  font-weight: 800;
+}
+
+html {
+  font-family: Avenir, sans-serif;
+  font-style: normal;
+}
+
+.search {
```
Viktor-Kost, line 20: It would be better to avoid margin collapsing ([article](https://www.freecodecamp.org/news/what-is-margin-collapse-and-how-to-avoid-it/).) 
So in this case you can just use margin-top for both form elements

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  width: 100%;
+  outline: none;
+  padding: 0 0 0 62px;
+  font-family: inherit;
+  height: 70px;
+  font-size: 16px;
+  font-weight: 800;
+  line-height: 22px;
+  background: #fff;
+  border: 1px solid #e1e7ed;
```
Viktor-Kost, line 48: Please make variables for colors you use several times and give clear names for these variables

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.button {
+  border: 0;
+  cursor: pointer;
+  background-color: transparent;
+  position: absolute;
+  margin-top: 25px;
```
Viktor-Kost, line 103: You have given an absolute position to buttons, so it would be better to use top/left/right/bottom rules to center this button instead of margins

Viktor-Kost (changes requested):
Good job! 

And, please, apply recommendation to make your code better.

Virass (changes requested):
Fix bugs from past comments

PR #2418, opened 2022-11-20
bnb

Viktor535x:


SanyaBratashchuk (changes requested):
Links is not working

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+@font-face {
+  font-family: "Avenir";
+  src: url(\fonts\Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(\fonts\Avenir-Heavy.ttf);
+  font-weight: bold;
+}
```
SanyaBratashchuk, line 15: check your fonts

src/index.html
```diff
@@ -11,6 +11,50 @@
...
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      class="form"
+      method="get"
+      action="/"
+      data-qa="big"
+    >
+      <div class="barContainer">
```
SanyaBratashchuk, line 20: why do you need this div, you can apply styles directly to form

src/style.css
SanyaBratashchuk: specify font-weigth for first 2

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+
+.icon {
+  position: absolute;
+  left: 26px;
+  top: 25px;
+  height: 19px;
+}
+
+.input {
+  font-family: "Avenir", sans-serif;
```
SanyaBratashchuk, line 40:
```suggestion
.input {
  font-family: inherit;
```

SanyaBratashchuk (changes requested):

PR #2427, opened 2022-11-21
search-bar was created

PolinaSukhorukova:
- [DEMO LINK](https://PolinaSukhorukova.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://PolinaSukhorukova.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,82 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir-Heavy;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,82 @@
...
+@font-face {
+  font-family: Avenir-Heavy;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: bold;
+}
+
+body {
+  margin-top: 20px;
+}
```
Anna-Tkachenko, line 17: will be better to set these styles using separate class instead of tag name

Anna-Tkachenko (approved):
Good job! 🔥

PR #2445, opened 2022-11-25
created search forms

barantarasNEW:
https://barantarasNEW.github.io/layout_search-bar-airbnb/
https://barantarasNEW.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <div class="container">
+      <form
+        class="form"
+        action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+        method="get"
+        spellcheck="false"
+        data-qa="big"
+      >
+        <label for="search" class="label">
+          <img
+            class="img"
+            src="/images/Search.svg"
+            alt="search icon"
+          >
+        </label>
+        <input
+          id="search"
+          class="input"
+          name="bigSearch"
+          type="text"
+          placeholder="Try &quot;Los Angeles&quot;"
+          data-qa="keypress"
+        >
+      </form>
+      <form
+        class="form form-small"
+        action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+        method="get"
+        spellcheck="false"
+        data-qa="small"
+      >
+        <label class="label form-small__label" for="search-small">
+          <img
+            class="img img-small"
+            src="/images/Search.svg"
+            alt="search icon"
+          >
+        </label>
+        <input
+          id="search-small"
+          class="input form-small__input"
+          name="smallSearch"
+          type="text"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </form>
+    </div>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,115 @@
+@font-face {
+  font-family: "Avenir";
+  src: url("fonts/Avenir-Book.ttf");
+  font-style: normal;
+  font-weight: 400;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url("fonts/Avenir-Heavy.ttf");
+  font-style: normal;
+  font-weight: 900;
+}
+
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+  font-size: 16px;
+}
+
+.container {
+  padding: 20px 8px;
+}
+
+.form {
+  width: 100%;
+  position: relative;
+}
+
+.form:not(:last-child) {
+  margin-bottom: 20px;
+}
+
+.label {
+  display: flex;
+  align-items: center;
+  height: 100%;
+  padding: 0 17px 0 26px;
+  position: absolute;
+}
+
+.img {
+  display: block;
+}
+
+.input {
+  width: 100%;
+  padding: 23px 26px 23px 62px;
+  font-family: "Avenir", sans-serif;
+  font-weight: 900;
+  font-size: 1rem;
+  line-height: 22px;
+  color: #3d4e61;
+  border: 1px solid #e1e7ed;
+  outline: none;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  transition: box-shadow 0.1s ease;
+}
+
+.input:hover {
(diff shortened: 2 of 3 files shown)


barantarasNEW:
image

src/style.css
```diff
@@ -0,0 +1,115 @@
...
+.form:not(:last-child) {
+  margin-bottom: 20px;
+}
+
+.label {
+  display: flex;
+  align-items: center;
+  height: 100%;
+  padding: 0 17px 0 26px;
+  position: absolute;
```
OlEzhA0, line 45: you should move `absolute` element with `top/bottom/right/left` properties, not paddings

src/style.css
```diff
@@ -0,0 +1,115 @@
...
+}
+
+.img {
+  display: block;
+}
+
+.input {
+  width: 100%;
+  padding: 23px 26px 23px 62px;
+  font-family: "Avenir", sans-serif;
```
OlEzhA0, line 55: `font-family: inherit` should work as well

OlEzhA0 (changes requested):
Hey! Good job! I left few comments, please check it and I can approve your work

src/style.css
barantarasNEW: done)

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  font-family: "Avenir";
+  src: url("fonts/Avenir-Heavy.ttf");
+  font-style: normal;
+  font-weight: 900;
+}
+
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
Viktor-Kost, line 18: Do not use other rules at `*` selector (besides box-sizing)

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  display: block;
+}
+
+.input {
+  width: 100%;
+  padding: 23px 26px 23px 62px;
+  font-family: inherit;
+  font-weight: 900;
+  font-size: 1rem;
+  line-height: 22px;
```
Viktor-Kost, line 59: Use not px values for line-height

src/index.html
```diff
@@ -11,6 +11,52 @@
...
+          id="search"
+          class="input"
+          name="bigSearch"
+          type="text"
+          placeholder="Try &quot;Los Angeles&quot;"
+          data-qa="keypress"
+        >
+      </form>
+      <form
+        class="form form-small"
```
Viktor-Kost, line 39: If you are using as modifiers use this syntax
```suggestion
        class="form form--small"
```

Apply changes all over the document

Viktor-Kost (changes requested):

barantarasNEW:
ok)

OlEzhA0:
Screenshot 2022-11-30 at 5 57 15 PM
remove outlines, check the design and apply changes according to it

src/style.css
```diff
@@ -0,0 +1,119 @@
...
+
+.img--small {
+  width: 11px;
+  height: 11px;
+}
+
+.form--small__input {
+  padding: 11px 13px 10px 33px;
+  font-size: 12px;
+  line-height: 19px;
```
Viktor-Kost, line 107: Use not px values for line-height. It is much better to use absolute values like 158% or 1.58

src/style.css
```diff
@@ -0,0 +1,119 @@
...
+form {
+  outline: none;
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+  font-size: 16px;
+
+  --border-color: #e1e7ed;
```
Viktor-Kost, line 28: It is better to use abstract naming, like "primary-color", "secondary-color", because you may want to use this color for background or other rules, not only for border

Viktor-Kost (changes requested):

barantarasNEW:
done )

src/index.html
```diff
@@ -11,6 +11,52 @@
...
+        <label class="label form--small__label" for="search-small">
+          <img
+            class="img img--small"
+            src="/images/Search.svg"
+            alt="search icon"
+          >
+        </label>
+        <input
+          id="search--small"
+          class="input form--small__input"
```
witflash, line 54: So, you have to add a modifier to the element name.
Like this:
```
"input input--small"
```
As you did for img for example

src/index.html
```diff
@@ -11,6 +11,52 @@
...
+        >
+      </form>
+      <form
+        class="form form--small"
+        action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+        method="get"
+        spellcheck="false"
+        data-qa="small"
+      >
+        <label class="label form--small__label" for="search-small">
```
witflash, line 45: And here `label label--small`

src/index.html
```diff
@@ -11,6 +11,52 @@
...
+          <img
+            class="img img--small"
+            src="/images/Search.svg"
+            alt="search icon"
+          >
+        </label>
+        <input
```
witflash, line 52: If you place `input` inside `label` you can avoid usage of `for` and `id` attributes. You may try this approach if you want

src/style.css
```diff
@@ -0,0 +1,118 @@
...
+input {
+  outline: none;
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+  font-size: 16px;
+
+  --primary-color: #e1e7ed;
```
witflash, line 28: A common best practice is to define custom properties on the [:root](https://developer.mozilla.org/en-US/docs/Web/CSS/:root) pseudo-class, so that it can be applied globally across your HTML document:

```
:root {
  --primary-color: #e1e7ed;
}
```

src/style.css
```diff
@@ -0,0 +1,118 @@
...
+
+.container {
+  padding: 20px 8px;
+}
+
+.form {
+  position: relative;
+}
+
+.form:not(:last-child) {
```
witflash, line 43: It's better to set `margin-bottom` to all elements where it's needed (e.g. for `.label`)

src/style.css
```diff
@@ -0,0 +1,118 @@
...
+
+.input::placeholder {
+  font-family: inherit;
+  font-weight: 300;
+  font-size: 1rem;
+  color: #3d4e61;
+  text-shadow: none;
+}
+
+.form--small__label {
```
witflash, line 95: `.label--small`

witflash (changes requested):
Overall good job, but pay attention to the comments.

witflash (changes requested):
Overall good job, but pay attention to the comments.

barantarasNEW:
done

AlisaVlasova (approved):

PR #2502, opened 2022-12-06
right solution with good DEMO-link

andrii-kurash:
https://andrii-kurash.github.io/layout_search-bar-airbnb/

Have a good day!


ArturErmolenko (changes requested):
Demo link doesn't work
Where Test Report link?)

PR #2531, opened 2022-12-07
add search bar

artem-sobolevskyi:
- [DEMO LINK](https://artem-sobolevskyi.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://artem-sobolevskyi.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,16 +1,44 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="UTF-8">
+    <meta charset="UTF-8" >
     <meta
       name="viewport"
       content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
     >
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,16 +1,44 @@
...
-  <body>
-    <h1>Search bar airbnb</h1>
+  <body class="body">
+    <form
+      action="#"
+      method="get"
+      class="search-form"
+      data-qa="big"
+    >
+      <div class="search">
```
Anna-Tkachenko, line 20: will be better to use label instead of simple div

src/style.css
```diff
@@ -0,0 +1,70 @@
+.body {
+  font-family: "Avenir", Arial, Helvetica, sans-serif;
```
Anna-Tkachenko, line 2: will be better to set font styles to html tag

Anna-Tkachenko (approved):
good job!

PR #2534, opened 2022-12-07
searchbar

Tsvylenyeva:
[DEMO LINK](https://Tsvylenyeva.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://Tsvylenyeva.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -20,7 +20,7 @@
         <img
           class="image"
           alt="search"
-          src="C:\nadiia-rusnak\projects\layout_search-bar-airbnb\src\src\images\Search.svg"
+          src="\images\Search.svg"
         >
         <input
           type="text"
@@ -31,14 +31,14 @@
     </div>
     <div class="container container-small">
       <form
-        action=" "
+        action="#"
         method="post"
         data-qa="big"
       >
         <img
           class="input image"
           alt="search"
-          src="C:\nadiia-rusnak\projects\layout_search-bar-airbnb\src\src\images\Search.svg"
+          src="\images\Search.svg"
         >
         <input
           type="text"
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,10 @@
 /* stylelint-disable color-hex-length */
+@import url("https://fonts.cdnfonts.com/css/avenir");
+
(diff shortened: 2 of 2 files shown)


src/index.html
```diff
@@ -11,6 +11,41 @@
...
+    <div class="container container-big">
+      <form
+        action=" "
+        method="post"
+        data-qa="big keypress"
+      >
+        <img
+          class="image"
+          alt="search"
+          src="C:\nadiia-rusnak\projects\layout_search-bar-airbnb\src\src\images\Search.svg"
```
Anna-Tkachenko, line 23: you should set relative path here for image (corresponding to the index.html file), not the absolute one

src/index.html
```diff
@@ -11,6 +11,41 @@
...
+        <input
+          type="text"
+          placeholder="Try &quot;Los Angeles&quot;"
+          class = "input input-big"
+        >
+      </form>
+    </div>
+    <div class="container container-small">
+      <form
+        action=" "
```
Anna-Tkachenko, line 34:
```suggestion
        action="#"
```

Anna-Tkachenko (changes requested):
Check path for images and I don't see any import for the needed fonts

src/style.css
```diff
@@ -0,0 +1,68 @@
+/* stylelint-disable color-hex-length */
+@import url("https://fonts.cdnfonts.com/css/avenir");
```
Anna-Tkachenko, line 2: right now your demo link has too long response time according to uploading fonts from this link
you're already have fonts folder in the project, just check in video how to add fonts via @font in correct way, and use existing fonts for your elements

src/style.css
```diff
@@ -0,0 +1,68 @@
...
+
+body {
```
Anna-Tkachenko, line 4: move font styles to html tag

Anna-Tkachenko (changes requested):
image
loop images should be inside of the inputs
if you'll have any questions how to do it, write please in channel for questions and let's solve this problem, or attend Q&A sessions

PR #2543, opened 2022-12-08
added search bar

vlad-liakh:


src/index.html
```diff
@@ -11,6 +11,48 @@
...
+          src="images/Search.svg"
+          alt="search field"
+          class="icon icon--small"
+        >
+
+        <input
+          placeholder="Try &quot;Los Angeles&quot;"
+          type="search"
+          name="search"
+          class="search smallSearch"
```
witflash, line 53: Use BEM here. If you need some modifier, you have to create modifier class: `class="search search--small"`

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+
+@font-face {
+  font-family: Avenir-Book;
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir-Heavy;
```
witflash, line 13: You can use simple `Avenir` font-family name here and above. So in CSS you can use `Avenir` name but with a defined "font-weight".

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 900;
+}
+
+body {
+  font-size: 16px;
+  font-family: Avenir-Book, Helvetica, sans-serif;
+  font-style: normal;
```
witflash, line 22: Do you really need `font-style: normal` here? I think it's the default in any browser.

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.form {
+  margin: 20px 0;
+  position: relative;
+}
+
+.search {
+  font-family: Avenir-Book, Helvetica, sans-serif;
```
witflash, line 31: Why do you need this rule here? All nested elements have to inherit font-family from "body" I think.
Can you check it?

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+  border-radius: 3px;
+  font-family: Avenir-Heavy, Helvetica, sans-serif;
+  color: #3d4e61;
+  outline: none;
+}
+
+.search::placeholder {
+  font-family: inherit;
+  font-weight: 300;
+  color: #3d4e61;
```
witflash, line 82: You can move the color to the [global variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties). It will be much better to follow [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) rules

witflash (changes requested):
Hey! Good job!
Take attention to the last comments.

src/style.css
vlad-liakh: Yes, you are right.

src/style.css
vlad-liakh: Yes, but if I explicitly state it font-family: inherit

src/style.css
vlad-liakh: ok

vlad-liakh:
Снимок экрана 2022-12-09 в 00 08 14

src/index.html
```diff
@@ -11,6 +11,48 @@
...
+        <input
+          placeholder="Try &quot;Los Angeles&quot;"
+          type="text"
+          name="search"
+          data-qa="keypress"
+          class="search"
+        >
+      </label>
+    </form>
+    <form
```
liliia-livandovskaya, line 36:
```suggestion
    </form>
    
    <form
```

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+}
+
+.search--small {
+  height: 42px;
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 19px;
+  font-weight: 400;
+
+}
```
liliia-livandovskaya, line 68:
```suggestion
  font-weight: 400;
}
```

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+
+.search:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+  font-family: Avenir-Heavy, Helvetica, sans-serif;
```
liliia-livandovskaya, line 77:
```suggestion
  font-family: inherit;
```

liliia-livandovskaya (approved):
Great job!

PR #2557, opened 2022-12-12
search bar was added

Daria13k:
https://Daria13k.github.io/layout_search-bar-airbnb/
https://Daria13k.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,45 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form 
+      class="form-search form-big"
+      action="#"
+      method="post"
+      data-qa="big"
+    >
+      <label class="form-label form-label-big">
+        <img
+          class="form-img form-img-big"
+          src="/images/Search.svg"
+          alt="search big img"
+        >
+        <input 
+          type="text"
+          class="form-input form-input-big"
+          placeholder="Try &quot;Los Angeles&quot;"
+          data-qa="keypress"
+        >
+      </label>
+    </form>
+
+    <form 
+      class="form-search form-small"
+      action="#"
+      method="post"
+      data-qa="small"
+      >
+      <label class="form-label form-label-small">
+        <img
+          class="form-img form-img-small"
+          src="/images/Search.svg"
+          alt="search small img"
+        >
+        <input 
+          type="text"
+          class="form-input form-input-small"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,108 @@
+* {
+  box-sizing: border-box;
+}
+
+@font-face {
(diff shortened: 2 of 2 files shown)


OlEzhA0:
Screenshot 2022-12-13 at 12 11 20 PM
remove blue outlines

src/index.html
```diff
@@ -11,6 +11,45 @@
...
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form 
+      class="form-search form-big"
+      action="#"
+      method="post"
+      data-qa="big"
+    >
+      <label class="form-label form-label-big">
```
OlEzhA0, line 20: label redundant

src/index.html
```diff
@@ -11,6 +11,45 @@
...
+        >
+      </label>
+    </form>
+
+    <form 
+      class="form-search form-small"
+      action="#"
+      method="post"
+      data-qa="small"
+      >
```
OlEzhA0, line 40:
```suggestion
    >
```

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+}
+
+.form-small {
+  width: 100%;
+  height: 42px;
+}
+
+.form-label {
+  position: relative;
+  display: block;
```
OlEzhA0, line 41: it's redundant

OlEzhA0 (changes requested):
Good job! I really like it.
Please check my comments and check what you can do

src/index.html
Daria13k: should I replace all 'labels' by <div>?

Daria13k: I tried to use 'outline:none' for input, also for 'form','form:focus','form:active', but there are no changes -the outline still exist and default value of my brauser is dark-grey. what do you recommend to do?

src/style.css
```diff
@@ -37,8 +37,6 @@ html {
 }
 
 .form-label {
-  position: relative;
-  display: block;
```
Virass: position: relative must be stay

Virass (changes requested):
Fix this bug

image

src/style.css
Daria13k: bug was fixed

src/index.html
witflash: Yep, if you have no label (no text) you may use `div`.

src/index.html
```diff
@@ -11,6 +11,45 @@
...
   <body>
-    <h1>Search bar airbnb</h1>
+    <form 
+      class="form-search form-big"
+      action="#"
+      method="post"
+      data-qa="big"
+    >
+      <label class="form-label form-label-big">
+        <img
```
witflash, line 21: You could use the image as a background for input. So you will have no need to use `img` tag. 
You can try it by yourself if you want.

witflash (approved):
Good job! 👍 
I left some recommendations for your future work.

src/index.html
Daria13k: Thank you for a recommendation, I will try!

PR #2569, opened 2022-12-15
search bar created

LinaDo1609:
- [DEMO LINK](https://LinaDo1609.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://LinaDo1609.github.io/layout_search-bar-airbnb/report/html_report/)

i have passed just 3 test from 6. I added options and settings from mockup, but i can't pass the test just because of .001px
maybe you can give me a piece of advice how to solve it


src/style.css
```diff
@@ -0,0 +1,106 @@
...
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir.ttc);
+  font-style: normal;
+  font-weight: 300;
+}
+
+body {
+  margin: 0;
+  padding: 20px 7.6px 7.5px ;
```
SanyaBratashchuk, line 24:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  padding-left: 62px;
+  border-radius: 4px;
+  width: 100%;
+  height: 70px;
+}
+
+.searchBarBig::placeholder {
+  font-family: "Avenir", sans-serif;
```
SanyaBratashchuk, line 41: Move font-family to the `html` tag selector and let inputs inherit it

SanyaBratashchuk (changes requested):
image

src/style.css
LinaDo1609: what does it mean?
can you explain your comment?

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+  font-weight: 300;
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+}
+
+body {
+  margin: 0;
+  padding: 20px 7.6px 7.5px;
```
SanyaBratashchuk, line 28: why do you need this decimal values?

SanyaBratashchuk (changes requested):
Try to implement it. If you have any questions feel free to ask in Slack 
image

src/style.css
LinaDo1609: Because i will fail test without this decimal values.

src/index.html
```diff
@@ -11,6 +11,39 @@
...
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+  
+    <form 
+      action="#"
+      method="get"
+      data-qa="big"
+    >
+      <div class="bigSearchBar">
```
OlEzhA0, line 20: why do we need this additional wrapper? it's redundant

src/style.css
```diff
@@ -0,0 +1,89 @@
...
+
+html {
+  font-family: Avenir, sans-serif;
+  font-style: normal;
+  font-weight: 300;
+}
+
+body {
+  margin-top: 0;
+  padding-top: 20px;
```
OlEzhA0, line 33: don't add padding-top to body
add margin-top to big input and it's enough

src/index.html
OlEzhA0: especially since you didn't use this class at all

OlEzhA0 (changes requested):
Hey! GJ
Please check my comments

src/style.css
LinaDo1609: i use padding in body to avoid margin collapse

src/index.html
LinaDo1609: maybe now i don't use it.
but as far as i know, a lot of developers say that it is better to make a div wrapper for each elements

LinaDo1609: can you explain why it is better to use margin top?

src/style.css
LinaDo1609: could you explain, why it is better to use margin-top?
is it a mistake to use padding in body?

src/index.html
```diff
@@ -11,6 +11,39 @@
...
+      data-qa="small"
+    >
+      <div class="smallSearchBar">
+        <input
+          class="input input--small"
+          type="search"
+          name="smallBar"
+          id="smallSearchBar"
+          
+          placeholder="Try &quot;Los Angeles&quot;"
```
ivanchuk-yurii, line 44:
```suggestion
          id="smallSearchBar"
          placeholder="Try &quot;Los Angeles&quot;"
```

src/index.html
ivanchuk-yurii: no, it's better to remove this, you already have outer `form` block

src/style.css
```diff
@@ -0,0 +1,89 @@
...
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-weight: 600;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir.ttc);
+  font-style: normal;
+  font-weight: 300;
+}
```
ivanchuk-yurii, line 23: you don't need this

src/style.css
ivanchuk-yurii: there will be no margin collapse bc set `margin-top: 0;` so it's better to remove `padding`

ivanchuk-yurii (changes requested):

SanyaBratashchuk (approved):

PR #2599, opened 2022-12-23
Search bar airbnb

danil-maksimenko:
- [DEMO LINK](https://danil-maksimenko.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://danil-maksimenko.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -34,7 +34,7 @@
       </label>
     </form>
     <form
-      class="search-form-small"
+      class="search-form search-form-small"
       method="get"
       action="/"
       data-qa="small"
@@ -51,7 +51,7 @@
           class="input input-small"
           type="text"
           name="search-small"
-          id="search-input-smal"
+          id="search-input-small"
           placeholder="Try &quot;Los Angeles&quot;"
         >
       </label>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,50 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
```
TarasHoliuk, line 14: Inputs should have some empty space from both sides:
image

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+.search-form {
+  position: relative;
+  margin-top: 20px;
+  height: 70px;
+}
+
+.search-form-small {
+  position: relative;
+  margin-top: 20px;
+}
```
TarasHoliuk, line 74: add class `search-form` to small form as well
and in `search-form-small` just overwrite styles which are different from the original (height in this case)

do the same everywhere it's needed

TarasHoliuk (changes requested):
Try to adjust styles to pass tests

nazarmatsevych (approved):

PR #2619, opened 2022-12-30
search form commit

tania-kuzmenko:
https://tania-kuzmenko.github.io/layout_search-bar-airbnb/
https://tania-kuzmenko.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,138 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: "Heavy";
+  font-weight: 800;
+}
+
+* {
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
+
+.search-form {
+  display: flex;
+  justify-content: center;
+  margin: 20px auto 0;
+}
+
+.search-form-big {
+  width: 100%;
+  height: 70px;
+}
+
+.search-form-small {
+  width: 100%;
+  height: 42px;
+}
+
+.big-label {
+  position: relative;
+  display: block;
+  width: 100%;
+}
+
+.small-label {
+  position: relative;
+  display: block;
+  width: 100%;
+}
+
+.big-label:after {
+  position: absolute;
+  content: "";
+  display: block;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 19px;
+  height: 19px;
+  top: 25px;
+  left: 26px;
+}
+
+.small-label:after {
+  position: absolute;
+  content: "";
+  display: block;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,138 @@
...
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: "Heavy";
+  font-weight: 800;
+}
+
+* {
+  padding: 0;
```
OlEzhA0, line 15: why do we need this?

src/style.css
```diff
@@ -0,0 +1,138 @@
...
+}
+
+.search-form {
+  display: flex;
+  justify-content: center;
+  margin: 20px auto 0;
+}
+
+.search-form-big {
+  width: 100%;
```
OlEzhA0, line 31: it's redundant style

src/style.css
```diff
@@ -0,0 +1,138 @@
...
+}
+
+.search-form-small {
+  width: 100%;
+  height: 42px;
+}
+
+.big-label {
+  position: relative;
+  display: block;
```
OlEzhA0, line 42: why do we need `display: block`?

src/style.css
```diff
@@ -0,0 +1,138 @@
...
+  width: 100%;
+}
+
+.small-label {
+  position: relative;
+  display: block;
```
OlEzhA0, line 48: the same

src/style.css
```diff
@@ -0,0 +1,138 @@
...
+}
+
+.search-input-small {
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 19px;
+}
+
+.search-input-big::placeholder {
+  font-family: "Avenir", sans-serif;
```
OlEzhA0, line 103: inherit?

src/style.css
```diff
@@ -0,0 +1,138 @@
...
+
+.search-input-big:default {
+  box-shadow: 0 1px 8px
+    rgba(61, 78, 97, 0.1);
+}
+
+.search-input-small:default {
+  box-shadow: 0 1px 8px
+    rgba(61, 78, 97, 0.1);
+}
```
OlEzhA0, line 120:
```suggestion
.search-input-big:default,
.search-input-small:default {
  box-shadow: 0 1px 8px
    rgba(61, 78, 97, 0.1);
}
```

src/style.css
```diff
@@ -0,0 +1,138 @@
...
+    rgba(61, 78, 97, 0.2);
+}
+
+.search-input-big:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
+
+.search-input-small:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
```
OlEzhA0, line 138: the same as above

OlEzhA0 (changes requested):
hey! good job! Please check my comments

src/style.css
tania-kuzmenko: it was indicated in Readme that "Remember that inputs and other interactive elements don’t inherit font styles by default." - I understand it like it is necessary to add this rule in placeholder  but  you are right it is inherited

PR #2635, opened 2023-01-03
Search bars created

khomenkom:
- [DEMO LINK](https://khomenkom.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://khomenkom.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,118 @@
+* {
+  padding: 0;
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 300;
+  font-style: normal;
+  src: url("./fonts/Avenir-Book.ttf");
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 600;
+  font-style: normal;
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
+body {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
+
+.searchbar {
+  margin: 20px auto 0;
+}
+
+.big-searchbar {
+  width: 100%;
+  height: 70px;
+}
+
+.small-searchbar {
+  width: 100%;
+  height: 42px;
+}
+
+.searchbar-label {
+  position: relative;
+  display: block;
+  width: 100%;
+}
+
+.searchbar-label:after {
+  position: absolute;
+  content: "";
+  display: block;
+}
+
+.searchbar-label-big:after {
+  background-image: url("./images/Search.svg");
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 19px;
+  height: 19px;
+  top: 25px;
+  left: 26px;
+}
+
+.searchbar-label-small:after {
+  background-image: url("./images/Search.svg");
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
+
+.input {
+  width: 100%;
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,118 @@
+* {
+  padding: 0;
```
vlad-demchuk, line 2: it's better not use reset styles for `*` selector
move it where you need

src/style.css
```diff
@@ -0,0 +1,118 @@
...
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 600;
+  font-style: normal;
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
+body {
+  font-family: "Avenir", sans-serif;
```
vlad-demchuk, line 21: apply fonts for `html` selector

src/style.css
```diff
@@ -0,0 +1,118 @@
...
+  margin: 20px auto 0;
+}
+
+.big-searchbar {
+  width: 100%;
+  height: 70px;
+}
+
+.small-searchbar {
+  width: 100%;
```
vlad-demchuk, line 35: don't repeat yourself
you could use one separate class and use set common styles there (`width: 100%;`)
upd you already have such class (`searchbar`)

vlad-demchuk: check all such cases

vlad-demchuk (changes requested):

src/style.css
khomenkom: Thank you for your feedback. Styles fixed.

src/style.css
```diff
@@ -0,0 +1,110 @@
+html {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+  padding: 0;
```
tetiana-ivanchenko, line 4: html hasn't default paddings, so move it to selector you need to reset

src/style.css
```diff
@@ -0,0 +1,110 @@
...
+  box-sizing: border-box;
```
tetiana-ivanchenko, line 5: move it to the * selector

tetiana-ivanchenko (changes requested):

src/style.css
vlad-demchuk: it meant move `box-sizing` to `*`

leave font styles in `html`

src/style.css
```diff
@@ -0,0 +1,110 @@
+* {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+  box-sizing: border-box;
+  padding: 0;
```
vlad-demchuk, line 5: and don't reset padding for `*` selector. it will be better to some special elements.

vlad-demchuk (changes requested):

src/style.css
khomenkom: Ok, I misunderstood. Thanks for the clarification, hope this time it's ok.

src/style.css
```diff
@@ -0,0 +1,112 @@
...
+  padding-left: 62px;
+  font-size: 16px;
+  line-height: 22px;
+}
+
+.small-input {
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 19px;
+}
```
tetiana-ivanchenko, line 91: you can move styles for big input, into the .input selector, and than modificate some values only for small input, the same for all classes,  it will reduce the number of classes

tetiana-ivanchenko (changes requested):

nazarmatsevych (approved):

PR #2661, opened 2023-01-08
new search bar was created

vlad-fil:
- [DEMO LINK](https://vlad-fil.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://vlad-fil.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,87 @@
+/* stylelint-disable font-family-no-missing-generic-family-keyword */
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+form {
+  position: relative;
+  margin-top: 20px;
+}
+
+[data-qa="big"] input {
+  box-sizing: border-box;
+  padding-left: 62px;
+  width: 100%;
+  height: 70px;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+
+  font-family: Avenir;
+  font-weight: bold;
+  font-size: 16px;
+  line-height: 22px;
+  color: #3d4e61;
+}
+
+[data-qa="small"] input {
+  box-sizing: border-box;
+  padding-left: 33px;
+  width: 100%;
+  height: 42px;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+
+  font-family: Avenir;
+  font-weight: bold;
+  font-size: 14px;
+  line-height: 19px;
+  color: #3d4e61;
+}
+
+input:hover {
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+input:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+}
+
+[data-qa="big"] input::placeholder {
+  font-family: Avenir;
+  font-weight: normal;
+  font-size: 16px;
+  line-height: 22px;
+  color: #3d4e61;
+}
+
+[data-qa="small"] input::placeholder {
+  font-family: Avenir;
+  font-weight: normal;
+  font-size: 14px;
+  line-height: 19px;
+  color: #3d4e61;
+}
+
+[data-qa="big"] img {
+  position: absolute;
+  left: 26px;
+  top: 25px;
+}
+
+[data-qa="small"] img {
+  position: absolute;
+  width: 11px;
+  height: 11px;
+  left: 13px;
+  top: 15px;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+form {
+  position: relative;
+  margin-top: 20px;
+}
+
+[data-qa="big"] input {
```
Viktor-Kost, line 12: remove this blue outline
Знімок екрана 2023-01-09 о 20 00 25

src/style.css
```diff
@@ -0,0 +1,87 @@
+/* stylelint-disable font-family-no-missing-generic-family-keyword */
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+form {
```
Viktor-Kost, line 7: it is better to use class selectors instead of tag ones 
Apple ALL OVER the page, please

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+
+  font-family: Avenir;
+  font-weight: bold;
+  font-size: 14px;
+  line-height: 19px;
+  color: #3d4e61;
+}
```
Viktor-Kost, line 44: You have a lot of repeatable styles. 

Following the BEM-methodology - use the `block` selector for the FIRST input (`class="input"`), where you apply common input styles

```
.input {
  {{common_input_styles}}
}
```

And then you apply `modifier` to the SECOND input (`class="input input--small"`) that have ONLY differences with base class input
```
.input--small {
  {{styles_with_difference}}
}
```
So in this way you follow DRY (Don't repeat yourself) principe

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+
+input:hover {
+  background: #fff;
+  border: 1px solid #e1e7ed;
```
Viktor-Kost, line 48: create variables for repeated colors

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  box-sizing: border-box;
+  padding-left: 62px;
+  width: 100%;
+  height: 70px;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+
+  font-family: Avenir;
```
Viktor-Kost, line 22: apply font-family to html/body selector and for input use just `font-family: inherit`;
Also, add `fallback` font

Viktor-Kost (changes requested):

src/style.css
```diff
@@ -0,0 +1,79 @@
+/* stylelint-disable font-family-no-missing-generic-family-keyword */
+@font-face {
+  font-family: Avenir;
```
Viktor-Kost, line 3: add fallback font here

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  --color-silver: #e1e7ed;
+  --color-gray: #3d4e61;
+  --color-dark: rgba(61, 78, 97, 0.1);
+}
+
+.form1,
+.form2 {
+  position: relative;
+  margin-top: 20px;
+}
```
Viktor-Kost, line 20: you did great with `.form__input` after recommendation, let's do the same with other elements. 
Common `.form` and modifier `.form--small`
Common `.form__image` and modifier `.form__image--small`
```suggestion
.form {
  position: relative;
  margin-top: 20px;
}
```

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+
+.form1 input::placeholder {
+  font-weight: normal;
+  color: var(--color-gray);
+}
+
+.form2 input::placeholder {
+  font-weight: normal;
+  color: var(--color-gray);
+}
```
Viktor-Kost, line 65: you have already have a selector for input 
Also these selectors are the same
```suggestion
.form_input::placeholder {
  font-weight: normal;
  color: var(--color-gray);
}
```

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  top: 25px;
+}
+
+.form2 img {
+  position: absolute;
+  width: 11px;
+  height: 11px;
+  left: 13px;
+  top: 15px;
+}
```
Viktor-Kost, line 79: create common .form__image and modifier .form__image--small

Viktor-Kost (changes requested):
Nice!!

It looks much better. Keep going

src/style.css
OlEzhA0: check this [guide](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md) (7 point)

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  --color-dark: rgba(61, 78, 97, 0.1);
+}
+
+.form {
+  position: relative;
+  margin-top: 20px;
+}
+
+.form__input {
+  box-sizing: border-box;
```
OlEzhA0, line 28: move this style to `*`  selector

OlEzhA0 (changes requested):
hey! good job!
please check my small comments

IvanFesenko (approved):
Good job

mykhalenych (approved):

PR #2674, opened 2023-01-11
Develop

Yevhenii0536:
- [DEMO LINK](https://yevhenii0536.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://yevhenii0536.github.io/layout_search-bar-airbnb/report/html_report/)


OleksiiBlezniuk (changes requested):
Please double check your solution with checklist, pay attention to code style

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 600;
```
OleksiiBlezniuk, line 23: Not critical, but why 600? `Heavy` font weight is usually 900 (can be 800)
600 - is something like `bold`

https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+}
+
+body {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
```
OleksiiBlezniuk, line 29: Usually we set font styles to `html` tag

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+
+.search-form {
+  display: grid;
```
OleksiiBlezniuk, line 32: I don't think `grid` is required here, just same `height` to input instead and you'll have same result

OleksiiBlezniuk (approved):

PR #2684, opened 2023-01-14
Develop

vladyslav-mosor:
https://vladyslav-mosor.github.io/layout_search-bar-airbnb/
https://vladyslav-mosor.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,42 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form action=" "
+      method="get"
+      class="search-form"
+      data-qa="big"
+    >
+      <label>
+        <div class="container">
+          <img src="images/Search.svg"
+            alt="search_icon"
+            class="icon"
+          >
+          <input type="text"
+            class="bar"
+            data-qa="keypress"
+            placeholder="Try &quot;Los Angeles&quot;"
+          >
+        </div>
+      </label>
+    </form>
+    <form action=" "
+      method="get"
+      class="search-form"
+      data-qa="small"
+    >
+      <label>
+        <div class="container">
+          <img src="images/Search.svg"
+            alt="search_icon"
+            class="icon icon--small"
+          >
+          <input type="text"
+            class="bar bar--small"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,42 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form action=" "
```
OleksiiBlezniuk, line 14: Action attribute shouldn't be empty, you can use `#` here

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
+
+body {
+  font-family: "Avenir", sans-serif;
```
OleksiiBlezniuk, line 16: Usually we set font related styles to `html` tag

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+  font-weight: 300;
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 18: Body don't have any paddings or borders, so there is no sense to use box-sizing here (it's not inheriting by default). Instead better to use `*` selector for this purpose

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+}
+
+.bar--small {
+  font-size: 14px;
+  line-height: 40px;
+}
+
+.bar-small:focus {
+  font-weight: 800;
+}
```
OleksiiBlezniuk, line 91: This rule is repeated, it's already set for `bar`

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+
+.bar-small:focus::placeholder {
+  font-weight: 300;
+  text-shadow: none;
+}
```
OleksiiBlezniuk, line 96: It's already set to `bar`, no need to duplicate

OleksiiBlezniuk (changes requested):
Please pay more attention to code style, double check your solution with checklist

PR #2715, opened 2023-01-22
Develop

ctk12:


polosanya (changes requested):
Check yourself with the design. 
1. The icon should be bigger
image
image
2. Font should be bolder when typing
image
image

qoa11a (approved):

PR #2722, opened 2023-01-23
Solution (new-pr)

superzuc:
https://superzuc.github.io/layout_search-bar-airbnb/

https://superzuc.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,13 +12,14 @@
   </head>
   <body>
     <div class="searchBar_big_container">
-      <form 
-        action="https://mate-academy-form-lesson.herokuapp.com/create-application" 
+      <form
+        action="https://mate-academy-form-lesson.herokuapp.com/create-application"
         method="post"
         data-qa="big"
-        >
+      >
         <label for="searchBar_big"></label>
-        <input 
+        <input
+          class="searchBar_big"
           type="search"
           id="searchBar_big"
           name="searchBar"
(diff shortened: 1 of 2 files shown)


superzuc:
This is a new project 'cause the link did not work so i've re-installed all npm in a new project and set npm run deploy. Now the first link works but the test link (the second one) doesn't. I ensure you that all the 6 tests are passed, it's all day long that i'm working on the problem. Thanks

src/index.html
```diff
@@ -11,6 +11,39 @@
...
+        >
+        <button type="submit" class="submit_searchButton"></button>
+      </form>
+    </div>
+
+    <div class="searchBar_small_container">
+      <form 
+        action="https://mate-academy-form-lesson.herokuapp.com/create-application" 
+        method="post"
+        data-qa="small">
```
polosanya, line 36: Same here

src/style.css
```diff
@@ -0,0 +1,103 @@
...
+}
+  
+.searchBar_small_container {
+  margin: 0 8px 10px;
+}
+  
+#searchBar_big {
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
```
polosanya, line 33: Pay attention to CSS properties order
https://9elements.com/css-rule-order/

src/style.css
```diff
@@ -0,0 +1,103 @@
...
+  border-radius: 4px;
+  padding-left: 62px;
+  height: 70px;
+  font-family: Avenir-Book, Arial, Helvetica, sans-serif;
+  font-size: 16px;
+  width: 100%;
+}
+  
+#searchBar_big:hover {
```
polosanya, line 42: It is better not to use id selectors. Use class selectors instead

polosanya (changes requested):
Almost done!

PR #2737, opened 2023-01-24
Develop

konichek1:
[DEMO LINK](https://konichek1.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://konichek1.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,5 +1,6 @@
 * {
   box-sizing: border-box;
+  font-family: "Avenir", sans-serif;
 }
 
 :root {
@@ -27,31 +28,27 @@
   background: #fff;
 }
 
-.big_input {
-  font-family: "Avenir", sans-serif;
-  height: 70px;
+.search_input {
   width: 100%;
-  padding-left: 62px;
-  font-size: 16px;
-  line-height: 22px;
   display: flex;
   border: var(--br-solid);
-  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
   border-radius: 4px;
   outline: none;
 }
 
+.big_input {
+  height: 70px;
+  padding-left: 62px;
+  font-size: 16px;
+  line-height: 22px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+}
+
 .small_input {
-  font-family: "Avenir", sans-serif;
   height: 42px;
-  width: 100%;
   padding-left: 33px;
   font-size: 14px;
   line-height: 19px;
-  display: flex;
-  border: var(--br-solid);
-  border-radius: 4px;
-  outline: none;
 }
 
 .icon {


src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  font-weight: 400;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+}
+
+.small_input:focus {
+  font-weight: 900;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
```
witflash, line 93: You can move color value to the CSS variable

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  top: 25px;
+}
+
+.small_icon {
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
+/* search bar has 3 state default, hover and focus */
```
witflash, line 66: Remove any comments. You don't need them

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 700;
+}
+
+.search1 {
```
witflash, line 19: Why is it `search1`? It may be just `search` isn't it?

witflash (changes requested):
Well done!
Check my comments, please.

src/style.css
konichek1: due to a merge error, I thought maybe if something changes, it will pass without errors, but still an error when merge develop into master

src/style.css
```diff
@@ -0,0 +1,100 @@
...
+  height: 42px;
+  width: 100%;
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 19px;
+  display: flex;
+  border: var(--br-solid);
+  border-radius: 4px;
+  outline: none;
+}
```
IvanFesenko, line 55: Some styles are same for both - Do not repeat your self. Create separate class for this, use modifire for difference
2023-01-25_16h09_38

src/style.css
```diff
@@ -0,0 +1,100 @@
...
+
+.search {
+  margin-top: 20px;
+  position: relative;
+  background: #fff;
+}
+
+.big_input {
+  font-family: "Avenir", sans-serif;
+  height: 70px;
```
IvanFesenko, line 32: `font-family` should be in `html` selector? for input should be `inherit`

IvanFesenko (changes requested):
Good job. Almost done. Now need some refactor. Please check comments

IvanFesenko (approved):
Well done

PR #2744, opened 2023-01-24
created inputs

valeriiayarzhemska:
- [DEMO LINK](https://valeriiayarzhemska.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://valeriiayarzhemska.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,77 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf) format("truetype");
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf) format("truetype");
+  font-style: bold;
+  font-weight: 900;
+}
+
+html {
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+}
+
+.big-form,
+.small-form {
+  margin-top: 20px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  width: 100%;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-color: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  font-family: inherit;
+  color: #3d4e61;
+}
```
OleksiiBlezniuk, line 37: No need to apply this styles for both classes. Create one class for form and then overlap sizes for small one

Big search bar - is a default element, and small searchbar - is a modified default element

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+.form {
+  height: 70px;
+  left: 48px;
+  top: 71px;
+  font-size: 16px;
+  line-height: 22px;
+  background-size: 19px;
+  background-position: top 25px left 26px;
```
OleksiiBlezniuk, line 46: nice idea 👍

OleksiiBlezniuk (approved):
GJ, but next time if you need to create two states of element - create class for main state, and then create modifier that will overlap styles for particular state

PR #2746, opened 2023-01-24
added search bars into index.html,added styles,changed readme

Lilu580:
- [DEMO LINK](https://Lilu580.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Lilu580.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+      method="post"
+      data-qa="big"
+      class="bigger__search search"
+    >
+      <label
+      for="biggerBar"
+      class="bigger__search-bar search-bar"
+      >
+        <img
+          src="images/Search.svg"
+          alt="search"
+          class="bigger__search-image image"
+        >
+        <input
+          type="text"
+          name="search"
+          id="biggerBar"
+          data-qa="keypress"
+          class="bigger__search-field field"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
+    <form
+      action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+      method="post"
+      data-qa="small"
+      class="smaller__search search"
+    >
+      <label
+      for="smallerBar"
+      class="smaller__search-bar search-bar"
+      >
+        <img
+          src="images/Search.svg"
+          alt="search"
+          class="smaller__search-image image"
+        >
+        <input
+          type="text"
+          name="search"
+          id="smallerBar"
+          class="smaller__search-field field"
+          placeholder="Los Angeles"
+        >
+      </label>
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,116 @@
+
+* {
+  margin: 0;
+  background: #fff;
+}
+
+input {
+  outline: none;
+}
+
+::placeholder {
+  font-family: avenirBook, sans-serif;
+  font-weight: 300;
+  color: #3d4e61;
+}
+
+@font-face {
+  font-family: AvenirBook;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: AvenirHeavy;
+  src: url(./fonts/Avenir-Heavy.ttf);
+}
+
+html {
+  font-family: AvenirBook, sans-serif;
+  font-weight: 300;
+}
+
+.search {
+  width: 100%;
+}
+
+.bigger__search {
+  padding: 20px 81px 20px 48px;
+}
+
+.smaller__search {
+  padding: 0 81px 0 48px;
+}
+
+.search-bar {
+  display: inline-block;
+  width: 100%;
+}
+
+.field {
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+      method="post"
+      data-qa="big"
+      class="bigger__search search"
```
OleksiiBlezniuk, line 18: i think it's more clear to set generic class first and then override it by modifier, not otherwise

src/index.html
```diff
@@ -11,6 +11,54 @@
...
+    >
+      <label
+      for="biggerBar"
+      class="bigger__search-bar search-bar"
+      >
```
OleksiiBlezniuk, line 23:
```suggestion
      <label
        for="biggerBar"
        class="bigger__search-bar search-bar"
      >
```

OleksiiBlezniuk: pay attention to indentations

src/index.html
```diff
@@ -11,6 +11,54 @@
...
+    <form
+      action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+      method="post"
+      data-qa="small"
+      class="smaller__search search"
+    >
+      <label
+      for="smallerBar"
+      class="smaller__search-bar search-bar"
+      >
```
OleksiiBlezniuk, line 48:
```suggestion
      <label
        for="smallerBar"
        class="smaller__search-bar search-bar"
      >
```

OleksiiBlezniuk: And no need to set `for` attribute to label if you wrapping input in label

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  src: url(./fonts/Avenir-Heavy.ttf);
+}
+
+body {
+  margin-top: 0;
+}
+
+input {
+  outline: none;
+}
```
OleksiiBlezniuk, line 17: Ok, looks like reseting of default styles, but still, better to use classes for styles

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+
+::placeholder {
+  font-family: AvenirBook, sans-serif;
+  color: #3d4e61;
+}
```
OleksiiBlezniuk, line 22: No, you need to change particular input placeholder

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  padding: 25px 0 23px 62px;
+  font-size: 16px;
+  line-height: 22px;
+}
+
+.bigger__search-field::placeholder {
+  font-weight: 300;
+  font-size: 16px;
+  line-height: 22px;
+}
```
OleksiiBlezniuk, line 69: Write all styles to generic class, and then override sizes for small input

Big input here - is like default state of input, and small - is a modified default input, smaller version of default input

OleksiiBlezniuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,77 @@
+@font-face {
+  font-family: AvenirBook;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: AvenirHeavy;
+  src: url(./fonts/Avenir-Heavy.ttf);
+}
```
OleksiiBlezniuk, line 9: Import one font, but different font-weights

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+body {
+  margin: 0;
+}
+
+.search-bar {
+  margin-left: 48px;
```
OleksiiBlezniuk, line 16: are you sure that it's a correct margin here?

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  width: 900px;
```
OleksiiBlezniuk, line 17: it should be 100%

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  top: 45px;
+  left: 26px;
+}
+
+.field {
+  width: 100%;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  font-family: AvenirHeavy, sans-serif;
```
OleksiiBlezniuk, line 33: Import one font family, but with different font weights
set font styles to `html` tag and then inherit it if it wasn't inherited by default (using `inherit` value)

OleksiiBlezniuk (changes requested):
Inputs should have 100% width
Screenshot 2023-01-27 at 18 26 57

src/style.css
```diff
@@ -0,0 +1,152 @@
...
+  width: 100%;
+  font-weight: 900;
+  color: inherit;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  outline: none;
+  padding: 25px 0 23px 62px;
+}
+
```
Andr1yK, line 117: why do we need those comments?)

Andr1yK (changes requested):

src/style.css
Lilu580: ahah,my bad

illiarerun (approved):

PR #2766, opened 2023-01-24
searchbar

kuzmi491:
https://kuzmi491.github.io/layout_search-bar-airbnb/
https://kuzmi491.github.io/layout_search-bar-airbnb/report/html_report/


src/style.css
```diff
@@ -0,0 +1,108 @@
+* {
+  box-sizing: border-box;
+  font-family: Avenir, sans-serif;
```
Andr1yK, line 3: do u need font-family, f.e. for script\canvas tag?)

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  width: 100%;
+  height: 100%;
+  color: #3d4e61;
+  font-size: 16px;
+  line-height: 22px;
+  padding-left: 62px;
+  outline: none;
+}
+
+.searchTextForSmall {
```
Andr1yK, line 51:
```suggestion
.searchText--small {
```

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+.button {
+  position: absolute;
+  width: 19px;
+  height: 19px;
+  left: 26px;
+  top: 25px;
+  cursor: pointer;
+}
+
+.buttonForSmall {
```
Andr1yK, line 69: the same here

src/style.css
Andr1yK: avoid `for` in class names, IMHO it is not useful)

Andr1yK (changes requested):
Something is wrong, why 2 bars with the same height?

image

Andr1yK (approved):

PR #2790, opened 2023-01-26
Searhbars was created

s3rg4sh:
https://s3rg4sh.github.io/layout_search-bar-airbnb/
https://s3rg4sh.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,86 @@
+@font-face {
+  font-family: Avenir;
+  font-style: normal;
+  src: url("./fonts/Avenir.ttc");
+}
+
+input {
+  outline: none;
+}
+
+.searchbar {
+  display: block;
+  width: 100%;
+  margin-bottom: 18px;
+  background: #fff;
+  border: 1px solid #e1e7ed;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,86 @@
...
+  font-weight: 300;
+  background-image: url("./images/Search.svg");
+  background-repeat: no-repeat;
+  padding: 0;
+}
+
+.search-big {
+  margin-top: 19px;
+  height: 70px;
+  font-size: 16px;
```
IvanFesenko, line 28: Do not hard code `height` use padding instead

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+
+input {
+  outline: none;
+}
+
+.searchbar {
+  display: block;
+  width: 100%;
+  margin-bottom: 18px;
+  background: #fff;
```
IvanFesenko, line 15: Remove margin from input, add it to form

IvanFesenko (changes requested):
Good job. Small changes required

AlisaVlasova (approved):

PR #2829, opened 2023-02-03
done

RuslanaSl:


etojeDenys (changes requested):
please add DEMO LINK & REPORT LINK

IvanFesenko (changes requested):
Demo link isn't working. If you have some troubles with deploy - ask about it in fe_chart

src/style.css
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.form {
+  margin: 20px auto 0;
+}
+
+input {
+  border: none;
+  outline: none;
+}
```
qoa11a, line 34: don't apply styles to tags. Use classes instead

qoa11a (approved):

PR #2831, opened 2023-02-03
Created and styled two serch field

HYPER2307:


src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  background: no-repeat url(./images/Search.svg) linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+  font-weight: 900;
+}
+
+.input__big {
+  height: 70px;
+  font-size: 16px;
+  line-height: 22px;
```
IvanFesenko, line 52: Do not hard code  height - use padding instead
`padding: 23px 0 23px 62px;`

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  background-position: 26px 25px;
+  background-size: 19px 19px;
+  padding-left: 62px;
+}
+
+.input__small {
+  height: 42px;
+  font-size: 14px;
```
IvanFesenko, line 60: Same here

IvanFesenko (approved):
Good job. For future do not hard code height where you can do not do it

PR #2850, opened 2023-02-09
add core searchbar

melonixxx:
[DEMO LINK](https://melonixxx.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://melonixxx.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,109 @@
+* {
+  padding: 0;
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 300;
+  font-style: normal;
+  src: url(fonts/Avenir-Book.ttf) format(truetype);
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 600;
+  font-style: normal;
+  src: url(fonts/Avenir-Heavy.ttf) format(truetype);
+}
+
+body {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
+
+.form {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,109 @@
+* {
+  padding: 0;
```
vlad-demchuk, line 2: don't reset paddings for all elements. just use it for some special tags that you need to reset

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 600;
+  font-style: normal;
+  src: url(fonts/Avenir-Heavy.ttf) format(truetype);
+}
+
+body {
```
vlad-demchuk, line 20:
```suggestion
html {
```

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+  box-shadow: 0 1px 8px 0 #3d4e611a;
+  outline: none;
+  font-weight: 300;
+  font-family: inherit;
+}
+
+.input_big {
+  padding-left: 62px;
+  height: 70px;
+}
```
vlad-demchuk, line 43: you could add these styles to `.input`, and only change styles for `.input_small`

vlad-demchuk: the same for other cases to avoid duplications

vlad-demchuk (changes requested):

etojeDenys (approved):
LGTM

PR #2853, opened 2023-02-10
search-bar

PashaMalyshkin:
https://PashaMalyshkin.github.io/layout_search-bar-airbnb/
https://PashaMalyshkin.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,4 +1,4 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,101 @@
...
+/* Reset css */
+* {
+  margin: 0;
+  padding: 0;
+  border: none;
+  box-sizing: border-box;
+  background: none;
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+}
```
liliia-livandovskaya, line 29: - always set font styles to the `html` tag
- in * tag only box-sizing can be used, anything else is forbidden
- margin and padding you can set to the body

src/index.html
```diff
@@ -11,6 +11,39 @@
...
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+      class="search-form"
+      method="get"
+      data-qa="big"
+    >
+      <button type="submit" class="search-form__button">
```
liliia-livandovskaya, line 20: it should not be a button, it's better to use `label`

liliia-livandovskaya (approved):
Well done 👍

PR #2878, opened 2023-02-18
Implemented search bar in two different size

Titech0:


nazarmatsevych (changes requested):
Add demo and test reports links to your Pull Request description

vlad-demchuk (approved):

PR #2897, opened 2023-02-20
task search-bar

NadieinOleh:
I have problem with tests

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,46 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      class="search-form"
+      action="#"
+      method="get"
+      data-qa="keypress"
+    >
+      <label class="search-input">
+        <img
+         class="icon"
+         src="https://raw.githubusercontent.com/NadieinOleh/layout_search-bar-airbnb/b502445e6909cc80a492852e0d464ccf5f15cbf9/src/images/Search.svg"
+         alt="loop-big"
+        >
+        <input
+          type="text"
+          class="input input__big"
+          name="bigInput"
+          placeholder="Try &quot;Los Angeles&quot;"
+          data-qa="big"
+        >
+      </label>
+    </form>
+    <form
+    action="/"
+    method="get"
+    class="search-form"
+    >
+      <label class="search-input">
+        <img
+          class="icon icon__small"
+          src="https://raw.githubusercontent.com/NadieinOleh/layout_search-bar-airbnb/b502445e6909cc80a492852e0d464ccf5f15cbf9/src/images/Search.svg"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,46 @@
...
+          placeholder="Try &quot;Los Angeles&quot;"
+          data-qa="big"
+        >
+      </label>
+    </form>
+    <form
+    action="/"
+    method="get"
+    class="search-form"
+    >
```
qoa11a, line 39:
```suggestion
    <form
      action="/"
      method="get"
      class="search-form"
    >
```

src/style.css
```diff
@@ -0,0 +1,118 @@
...
+  font-weight: 300;
+  font-style: normal;
+  src: url(./fonts/Avenir.ttc);
+}
+
+/* body {
+  font-family: "Avenir", sans-serif;
+  font-style: normal;
+  font-weight: 300;
+} */
```
qoa11a, line 26: remove comments

qoa11a (changes requested):
you should remove default outline on focus
image

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.icon {
+  position: absolute;
+  top: 25px;
+  left: 26px;
+  height: 19px;
+  width: 19px;
+  display: block;
+
```
anstsot, line 87: remove empty line please

anstsot (changes requested):
Looks good but you have 2 failed tests in the test report. could you check them please?

src/style.css
NadieinOleh: I tried many times but it was unsuccessful with test

etojeDenys (commented):
please, if you have some problems, feel free to ask about them in the slack chat

polosanya (changes requested):
1. Input should take the whole height. 
image
2.  Check yourself with the design. Font size looks too small
image
image
3. There are also no styles for focus
image

src/style.css
```diff
@@ -0,0 +1,100 @@
...
+  font-style: normal;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+.search-form {
+  margin-top: 20px;
+}
+
+.input {
+  font-family: "Avenir", serif;
```
anstsot, line 20: it would be better to add font-family to HTML as separate style
```
html {
  font-family: Avenir, sans-serif;
}
```
and here you can just write `font-family: inherit;`. please check other places as well

anstsot (changes requested):
Looks good but you have 1 failed test in the report. could you please check?

NadieinOleh:
https://nadieinoleh.github.io/layout_search-bar-airbnb/report/html_report/

anstsot (approved):
Great job, approved!

PR #2899, opened 2023-02-20
searching form

MylaCrystal:
- [DEMO LINK](https://MylaCrystal.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://MylaCrystal.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,38 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="https://mate-academy-form-lesson.herokuapp.com/create-application"
+      method="get"
+      data-qa="keypress"
+    >
+      <img
+        class="bigSearchImg"
+        src="images/Search.svg"
+        alt="search"
+      >
+      <input
+        class="bigSearch searching"
+        type="search"
+        name="bigSearch"
+        data-qa="big"
+        placeholder="Try &quot;Los Angeles&quot;"
+      >
+    </form>
+
+    <form action="https://mate-academy-form-lesson.herokuapp.com/create-application" method="get">
+      <img
+        class="smallSearchImg"
+        src="images/Search.svg"
+        alt="search"
+      >
+      <input
+        class="smallSearch searching"
+        type="search"
+        name="smallSearch"
+        data-qa="small"
+        placeholder="Try &quot;Los Angeles&quot;"
+      >
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,73 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
+
+.searching {
+  width: 100%;
+  margin-top: 20px;
+  color: #3d4e61;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  font-family: Avenir, sans-serif;
+  font-size: 16px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  position: absolute;
+}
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,38 @@
...
+      method="get"
+      data-qa="keypress"
+    >
+      <img
+        class="bigSearchImg"
+        src="images/Search.svg"
+        alt="search"
+      >
+      <input
+        class="bigSearch searching"
```
liliia-livandovskaya, line 25: [checklist p.2](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md)

you can leave just `searching` class there. Also don't forget to move styles

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+  font-weight: 800;
+}
+
+.searching {
+  width: 100%;
+  margin-top: 20px;
+  color: #3d4e61;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  font-family: Avenir, sans-serif;
```
liliia-livandovskaya, line 21: you can set the front to the `html` tag

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+  font-size: 14px;
+  top: 82px;
+  left: 0;
+}
+
+.bigSearchImg {
+  width: 19px;
+  position: absolute;
+  top: 37px;
+  left: 26px;
```
liliia-livandovskaya, line 58: you do not need to use `top` and `left` there. use margins instead

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+  z-index: 1;
+}
+
+.smallSearchImg {
+  width: 11px;
+  position: absolute;
+  top: 117px;
+  left: 13px;
```
liliia-livandovskaya, line 66: the same here

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+  z-index: 1;
+}
+
+form {
```
liliia-livandovskaya, line 70: checklist p.1  Get used to style all elements using classes

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+.searching:focus {
+  outline: none;
+  border-radius: 3px;
+  font-weight: 800;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
+
+.bigSearch {
+  height: 70px;
+  top: -8px;
```
liliia-livandovskaya, line 40: don't use `top`

liliia-livandovskaya (changes requested):
Good job! Let's work on improvements

src/index.html
MylaCrystal: I have a question about this, wouldn't it be a problem to set the styles for both elements first and then change the styles again for the small search bar?

Or during display for the browser will be a greater load to add to the elements those styles that are unique for each search bar?

src/style.css
MylaCrystal: border styles I'll not move, right?

src/style.css
MylaCrystal: don't use at all or change on margin-top? 
if don't use it, I have this problem:
image

src/style.css
liliia-livandovskaya: right

src/index.html
liliia-livandovskaya: no, that would not be a problem. Try and you will see how it works in practice

src/style.css
```diff
@@ -0,0 +1,74 @@
...
+  width: 100%;
+  margin-top: 20px;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  position: absolute;
+  height: 70px;
+  padding-left: 62px;
+  font-size: 16px;
+  font-family: Avenir, sans-serif;
```
liliia-livandovskaya, line 38: always set font styles to the `html` tag and inherit where needed

liliia-livandovskaya (approved):
👍

src/style.css
MylaCrystal: it doesn't work in html tag

PR #2932, opened 2023-02-21
add two search bars

Anna-Kuzmenko:


src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+}
+
+.img {
+  position: absolute;
+  left: 26px;
+  top: 25px;
+  /* display: block; */
```
liliia-livandovskaya, line 80: dont forget to remove comments

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  font-weight: 300;
+}
+
+.search__form {
+  margin-top: 20px;
+  position: relative;
+}
+
+.search__input {
+  font-family: Avenir, sans-serif;
```
liliia-livandovskaya, line 30: always set font to `html` tag and inherit it where needed

liliia-livandovskaya (approved):
Great job!

PR #2951, opened 2023-02-22
search-bar done

stnayn:
- [DEMO LINK](https://stnayn.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://stnayn.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,91 @@
...
+.icon {
+  position: absolute;
+  left: 26px;
+  top: 25px;
+}
+
+.searchInput {
+  display: inline-block;
+  width: 100%;
+  font-family: "Avenir", sans-serif;
```
OleziO, line 34: Move this style to `body` selector

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
```
OleziO, line 36: Use HEX instead RGBA

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+  border-radius: 4px;
+  padding: 24px 25px 23px 62px;
+  font-size: 16px;
+  line-height: 21px;
+  color: #3d4e61;
```
OleziO, line 41: Add CSS variables for repeatable colors.

src/style.css
OleziO: Check other similar cases.

OleziO (changes requested):

src/style.css
stnayn: These are styles for input. If I move them my layout will break

anstsot (approved):
Great job! You can add small improvement with `font-family` style

src/style.css
stnayn: ok, thanks =)

PR #2963, opened 2023-02-23
Complete search bar

IvanBerdnyk:
- [DEMO LINK](https://ivanberdnyk.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ivanberdnyk.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,88 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+  font-style: normal;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 900;
+  font-style: normal;
+}
+
+html {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,89 @@
...
+
+.search_icon--big {
+  height: 19px;
+  width: 19px;
+  padding-left: 26px;
+  padding-top: 25px;
+
+}
+
+.search_icon--small {
```
vlad-demchuk, line 36: check this checklist point
image

vlad-demchuk (changes requested):
something wrong with the icon when I click on the field
image

src/style.css
IvanBerdnyk: image
image
Please need help!!!
When I run tests through npm, everything is okay, but when the tests are run on GitHub, it gives an error of size difference in small search bar

IvanBerdnyk: Problem solved! tnx

nazarmatsevych (approved):

PR #2980, opened 2023-03-01
create search bar

anna-chovnova:
- [DEMO LINK](https://anna-chovnova.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://anna-chovnova.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -5,6 +5,14 @@
   font-style: normal;
 }
 
+html {
+  font-family: Avenir, Arial, sans-serif;
+}
+
+* {
+  box-sizing: border-box;
+}
+
 .search-form {
   position: relative;
   margin-top: 20px;
@@ -17,21 +25,13 @@
 }
 
 .icon-small {
-  position: absolute;
   width: 11px;
   top: 15px;
   left: 13px;
 }
 
-.input {
-  box-sizing: border-box;
-}
-
 input::placeholder {
-  /* font-family: Avenir, Arial, sans-serif;
-  font-size: 16px; */
   color: #3d4e61;
-  /* font-style: inherit; */
 }
 
 input:hover {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,75 @@
...
+
+.input {
+  box-sizing: border-box;
+}
+
+input::placeholder {
+  /* font-family: Avenir, Arial, sans-serif;
+  font-size: 16px; */
+  color: #3d4e61;
+  /* font-style: inherit; */
```
IvanFesenko, line 34: remove comments

src/style.css
```diff
@@ -0,0 +1,75 @@
...
+.icon-small {
+  position: absolute;
+  width: 11px;
+  top: 15px;
+  left: 13px;
+}
+
+.input {
+  box-sizing: border-box;
+}
```
IvanFesenko, line 28: move `box-sizing` to `*` selector

src/style.css
```diff
@@ -0,0 +1,75 @@
...
+  height: 42px;
+  border: 1px solid rgba(225, 231, 237, 1);
+  border-radius: 4px;
+  width: 100%;
+  padding-left: 33px;
+
+  font-family: Avenir, Arial, sans-serif;
+  font-size: 14px;
+  color: #3d4e61;
+}
```
IvanFesenko, line 72: if you have two or more similar elements with portions of similar styles with different values - use one of the elements as the basic case, and override necessary styles for other cases. Explanation: The point is not in the names of the classes, the point is: when there are several similar elements, ex., 2 inputs, for one we can give a class input, for example, and for the second - input input--small. We write all the styles for .input, but for .input-small we write only those styles that differ in design, and we need this second input to look a little different. Element with class .input without extra classes should also look like a full-fledged styled element.
 
image

IvanFesenko (changes requested):
You should remove outline on focus state
2023-03-01_15h31_40
Also check comments

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+  font-weight: bold;
+  outline: none;
+}
+
+.input {
+  font-family: inherit;
+  height: 70px;
+  border: 1px solid rgba(225, 231, 237, 1);
```
OleziO, line 54: Use HEX instead of RGBA

OleziO (approved):

PR #2987, opened 2023-03-02
finished the task

EvhenIvchenko:
- [DEMO LINK](https://EvhenIvchenko.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://EvhenIvchenko.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,76 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Heavy.ttf);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+.icon-small {
+  top: 15px;
+  left: 13px;
+  width: 11px;
+  height: 11px;
+}
+/* EACH INPUT */
+.input {
+  display: block;
+  font-family: "Avenir" , sans-serif;
```
etojeDenys, line 40: we can move it to `html` selector

etojeDenys (approved):
almost perfect 🤩

src/style.css
EvhenIvchenko: You mean font family?
Inputs doesn't inherit font properties from parents.

PR #2991, opened 2023-03-03
created search bar

vintwp:
https://<your_account>.github.io/layout_search-bar-airbnb/
https://<your_account>.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,37 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form action="/"
+          method="GET"
+          class="form-search"
+          data-qa="big"
+    >
+      <div class="form-field">
+        <img  src="images/Search.svg"
+              alt="Search form icon"
+              class="search-icon">
+        <input  type="text"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,37 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form action="/"
```
slavanosatskyi, line 14: Please, check your code style 
Screenshot 2023-03-03 at 17 52 56

slavanosatskyi (changes requested):
nice work, though code style should be fixed

src/index.html
vintwp: Thanks for your quck review. I've fixed code style in html. But I have one question - why did test in github not complete? Test shows, that is different in pixels? I couln't see, what difference between my work and reference? Thanks!

slavanosatskyi (approved):
in general everything is fine with your tests https://vintwp.github.io/layout_search-bar-airbnb/report/html_report/.

probably some issue with tests themselves

PR #2998, opened 2023-03-05
create search form

Solovei-Andrii:
Solovei-Andrii.github.io/layout_search-bar-airbnb/
https://Solovei-Andrii.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,79 @@
+.input {
+  width: 100%;
+  margin-bottom: 20px;
+  font-weight: 300;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  color: #3d4e61;
+  font-family: Avenir, sans-serif;
+}
+
+form:first-child {
+  margin-top: 20px;
+}
+
+form:last-child {
+  margin-bottom: 0;
+}
+
+.big_search {
+  background-image: url("./images/Search.svg");
+  background-position: left 25px top 24px;
+  background-repeat: no-repeat;
+  padding-left: 62px;
+  background-size: 19px 19px;
+  height: 70px;
+  line-height: 22px;
+  box-sizing: border-box;
+  font-size: 16px;
+}
+
+.small_search {
+  background-image: url("./images/Search.svg");
+  background-position: left 12px top 14px;
+  background-repeat: no-repeat;
+  padding-left: 33px;
+  background-size: 11px 11px;
+  height: 42px;
+  box-sizing: border-box;
+  font-size: 14px;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: "Book";
+  font-weight: 300;
+}
+
+@font-face {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,79 @@
+.input {
+  width: 100%;
+  margin-bottom: 20px;
+  font-weight: 300;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  color: #3d4e61;
+  font-family: Avenir, sans-serif;
```
vlad-demchuk, line 9: move `font-family` to `html` and use `inherit`

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+}
+
+form:first-child {
+  margin-top: 20px;
+}
+
+form:last-child {
```
vlad-demchuk, line 16: don't style tags

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+
+.small_search {
+  background-image: url("./images/Search.svg");
+  background-position: left 12px top 14px;
+  background-repeat: no-repeat;
+  padding-left: 33px;
+  background-size: 11px 11px;
+  height: 42px;
+  box-sizing: border-box;
+  font-size: 14px;
```
vlad-demchuk, line 40: DRY
image

vlad-demchuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  background-position: left 25px top 24px;
+  
+  padding-left: 62px;
+  background-size: 19px 19px;
+  height: 70px;
+  line-height: 22px;
+  font-size: 16px;
+}
+
+.small_search {
```
vlad-demchuk, line 38: don't add extra classes

make it like that
```
.search {
  all styles
}

.small_search {
  styles that differ from `.search`
}

Comment (vlad-demchuk) [reply src/style.css:38]: no need to use small_search and big_search

**Review by vlad-demchuk — CHANGES_REQUESTED** [review]

File: src/style.css (hunk 4 above)
Comment (vlad-demchuk) [reply src/style.css:38]: image

File: src/style.css [hunk 5]
```diff
@@ -0,0 +1,76 @@
...
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: "Heavy";
+  font-weight: 800;
+}
+
+.big_search:hover {
+  font-style: "Book";
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
```
nazarmatsevych, line 61: `big_search` class looks redundant, you can move this styles under general class `search`

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+
+.form:last-child {
+  margin-bottom: 0;
+}
+
+.search {
+  background-image: url("./images/Search.svg");
+  background-repeat: no-repeat;
+  box-sizing: border-box;
```
nazarmatsevych, line 27: move box-sizing to * selector

nazarmatsevych (changes requested):

PR #3014, opened 2023-03-07
add searchbars

Zibi95:
[DEMO LINK](https://Zibi95.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://Zibi95.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,7 +12,12 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <form class="search-form" data-qa="big">
+    <form
+      action="/"
+      method="GET"
+      class="search-form"
+      data-qa="big"
+    >
       <label for="search" class="search-form__label">
         <img
           src="./images/Search.svg"
@@ -32,7 +37,12 @@
       >
     </form>
 
-    <form class="search-form search-form--size_small" data-qa="small">
+    <form
+      action="/"
+      method="GET"
+      class="search-form search-form--size_small"
+      data-qa="small"
+    >
       <label
         for="search--size_small"
         class="search-form__label search-form__label--size_small"
--- a/src/index.html
+++ b/src/index.html
@@ -11,58 +11,60 @@
     <link rel="stylesheet" href="./reset.css">
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body class="page">
-    <form
-      action="/"
-      method="GET"
-      class="search-form page__search-form"
-      data-qa="big"
-    >
-      <label for="search" class="search-form__label">
-        <img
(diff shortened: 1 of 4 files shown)


DorotaLeniecDev (changes requested):

src/index.html
```diff
@@ -11,6 +11,43 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="search-form search-form-big" data-qa="big">
```
DorotaLeniecDev, line 14: According to task description please provide action and method for forms

- You have to put input inside the form for correct data processing. Form should have action and method attributes.

DorotaLeniecDev (changes requested):

Zibi95:
I couldn't add `gap` because puppeter tests doesn't see that property. 
Here is Tip and Hints from other assigment:
> Don't use gap property for indents. It's not supported in tests, use margin instead.

Radoslaw-Czerniawski:
A few comments from my side: 
- Generally avoid setting margins on both `top` and `bottom`, when positioning elements on top of the other. In set of elements that are stacked, separate them by, for example setting `margin` only for `top`. This way you will always know that next element moves away from the previous according to it's own styles, rather than the previous elements styles cause the next one to change position
- Also top and bottom margins cause your forms margins to collide. They are separated not by combined margins value (which is fine) however it's confusing, if one margin was bigger than the other, bigger value would win. 
- You have to position your cursor in the middle of the form to focus input. This may be kinda confusing for the user that's clicking on what seems to be the input but it does not focus. I think setting on your input `width: 100%` to cover parents height will work.

src/index.html
```diff
@@ -11,11 +11,11 @@
     <link rel="stylesheet" href="./reset.css">
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
+  <body class="page">
```
Radoslaw-Czerniawski, line 14: Don't set classes on body element. This is redundant as there is no possibility for you to have more than one `body` element. If you must target body in css do it with tag, however I assume you want to have some type of container for these forms, I suggest wrapping them in a div with a class `wrapper`.

Radoslaw-Czerniawski (changes requested):

DorotaLeniecDev (approved):
Thanks for applying all the suggestions. Now task meet all the requirements :)
Sorry I have missed **gap** issue.

Radoslaw-Czerniawski:
Very good work with margins, input height and overall this whole PR 😄

Radoslaw-Czerniawski (approved):

PR #3027, opened 2023-03-07
Develop

GlibVK:
I hope i delete all duplicate.)))

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,7 +12,7 @@
   </head>
 
   <body>
-    
+
     <form
       action="big"
       method="get"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -10,7 +10,52 @@
...
+
+      <img
+        src="./images/Search.svg"
+        alt="magnifying glass"
+        class="icon icon_big"
+      >
+
+      <input
+        type="text"
+        class="input input__big"
```
agniya-i, line 31:
```suggestion
        class="input"
```
Please follow the approach from the [checklist](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md) p2. It would be better to create one common class, and just override properties for small input. Same for the other elements

agniya-i (changes requested):

PR #3033, opened 2023-03-08
add search bars

MartaChobaniuk:
- [DEMO LINK](https://martachobaniuk.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://martachobaniuk.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,114 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 900;
+}
+
+.search-form {
+  position: relative;
+}
+
+.input {
+  box-sizing: border-box;
+  position: absolute;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 70px;
+  margin: 20px 0;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  padding-left: 62px;
+}
+
+.input::placeholder {
+  font-family: Avenir, sans-serif;
+  font-style: normal;
+  font-weight: 300;
+  font-size: 16px;
+  line-height: 22px;
+  color: #3d4e61;
+}
+
+.input:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.input:focus {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  font-style: normal;
+  font-weight: 900;
+}
+
+.search-form {
+  position: relative;
+}
+
+.input {
+  box-sizing: border-box;
```
sTorba24, line 20: move it to html or body

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  color: #687c94;
+}
+
+.search-form-small {
+  position: relative;
+  top: 90px;
+}
+
+.input-small {
+  box-sizing: border-box;
```
sTorba24, line 77: remove it

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  width: 100%;
```
sTorba24, line 78: redundant, you already set it for .input

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+
+.search-form {
+  position: relative;
+}
+
+.input {
+  box-sizing: border-box;
+  position: absolute;
+  left: 0;
+  top: 0;
```
sTorba24, line 23: not sure you need to use position absolute for inputs

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  width: 11px;
+  height: 11px;
+  left: 13px;
+  top: 36px;
+  color: #687c94;
+  z-index: 2;
+}
+
+.input-small:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
```
sTorba24, line 102: don't duplicate styles, you already have it for .input

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  border-radius: 4px;
+}
+
+.input-small:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
sTorba24, line 107: the same, duplicating

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  border-radius: 3px;
+  font-family: Avenir, sans-serif;
```
sTorba24, line 109: set it for html and use 'inherit' value

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  font-style: normal;
```
sTorba24, line 110: it's default value, you can remove

sTorba24 (changes requested):
I left you some comments that need to be fixed
Also main thing is to avoid duplicating code, if you have input and smaller input set ALL styles for input add it as a class to big and small and then use class like `input-small` where you can just rewrite styles that are different between big-small

PR #3034, opened 2023-03-08
searchbar task

bjeziorek:
- [DEMO LINK](https://bjeziorek.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://bjeziorek.github.io/layout_search-bar-airbnb/report/html_report/)


michalcywka (changes requested):
- Border should be different when focusing,
- Placeholder should not change font style when focused
- The borders of the input are invisible

michalcywka (changes requested):

michalcywka (changes requested):
Filter should not affect placeholder, checklist 12

michalcywka (approved):

PR #3036, opened 2023-03-08
make searchbar

Ozwiena:
- [DEMO LINK](https://ozwiena.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ozwiena.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,42 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="https://www.google.com"
+      method="get"
+      class="form"
+    >
+      <div class="div" data-qa="big">
+        <img
+           class="image"
+           src="images/Search.svg"
+           alt="Search"
+        >
+        <input
+          class="big bar"
+          type="search"
+          placeholder="Try &quot;Los Angeles&quot;"
+          data-qa="keypress"
+        >
+      </div>
+    </form>
+    <form
+      action="https://www.google.com"
+      method="get"
+      class="form"
+    >
+      <div data-qa="small" class="div">
+        <img
+          class="image image-small"
+           src="images/Search.svg"
+           alt="Search"
+        >
+        <input
+          type="search"
+          class="bar bar-small"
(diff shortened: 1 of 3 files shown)


michalcywka (changes requested):
We still need hover and focus effects :)

michalcywka (changes requested):
I still don't see any hover effects in the Demo Link. Also there should be some styling for the text when input is in `focus` state.

src/style.css
```diff
@@ -0,0 +1,71 @@
+@font-face {
```
michalcywka, line 1: Checklist point 7

michalcywka (changes requested):

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf),
+    url(fonts/Avenir-Heavy.ttf);
```
michalcywka, line 4: This is almost good, but each font-face should have only one `url` :)

src/index.html
```diff
@@ -11,6 +11,42 @@
...
+    <form
+      action="https://www.google.com"
+      method="get"
+      class="form"
+    >
+      <div data-qa="small" class="div">
+        <img
+          class="image image-small"
+           src="images/Search.svg"
+           alt="Search"
```
michalcywka, line 42: Something off with the indentation of the attributes :)

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  border-radius: 3px;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  font-weight: bold;
+}
+
+.bar-small:focus {
+  font-weight: 900;
+}
+
+::placeholder {
```
michalcywka, line 75: Limit it to a particular class :)

michalcywka (changes requested):
Some small changes and should be good :)

michalcywka (changes requested):

michalcywka (approved):

PR #3037, opened 2023-03-08
v.0000000000001

ctgdprzyb:


michalcywka (changes requested):
Please look at the mockup, styles are different there :)

michalcywka (changes requested):
Gradient is not applied on focus
Also checklist point 7, see weights

michalcywka (changes requested):
The placeholders have different styles.

michalcywka (approved):

PR #3038, opened 2023-03-08
add task solution

marshalek37:
[DEMO LINK](https://marshalek37.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://marshalek37.github.io/layout_search-bar-airbnb/report/html_report/)


michalcywka (changes requested):
Please look at the mockup, for example hover colours are different there :)

michalcywka (commented):
Checklist point 12.
Also styles for focus are different than in the mockup, please change it :)

michalcywka (changes requested):
Placeholder shouldn't disappear until we type something in :)
Also please don't revert `.linthtmlrc.json` and `package.json`

PR #3049, opened 2023-03-10
add task solution

hsvirina:
https://hsvirina.github.io/layout_search-bar-airbnb/
https://hsvirina.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,82 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
+
+body {
+  font-family: "Avenir", Helvetica, sans-serif;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.form {
+  margin-top: 20px;
+}
+
+.input {
+  width: 100%;
+  font-family: inherit;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  outline: none;
+  font-weight: 300;
+  height: 70px;
+  padding-left: 62px;
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
+  background-position: 26px 25px;
+  font-size: 16px;
+  line-height: 22px;
+}
+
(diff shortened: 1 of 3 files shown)


hsvirina:
And question.. Firstly I tried to do this task with 2 divs, first was for img, second was for input. But I could not do it correctly. That's why I have made decision to do it through backgroung for img. But still I want to try to do it with divs for o figure out it. How can I do it for you check after then my code?

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
+
+body {
+  font-family: "Avenir", Helvetica, sans-serif;
```
etojeDenys, line 16: move it into html selector

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+
+.input:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.input:focus {
+  background: url(./images/Search.svg) 26px center / 19px no-repeat, linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border-radius: 3px;
+  font-weight: 900;
+  color: #3d4e61;
```
etojeDenys, line 62: let`s move this color into variable and then reuse it everywhere

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+  filter: drop-shadow(0 4px 4px rgba(0, 0, 0, 0.25));
+}
+
+.input_small:focus {
+  background: url(./images/Search.svg) 13px 15px no-repeat, linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
etojeDenys, line 67: try to add this Search.svg icon as ::before pseudo element or img tag

etojeDenys (changes requested):
almost done

src/style.css
hsvirina: What do you mean 'img tag'? To create <div> with  inside?

src/style.css
```diff
@@ -0,0 +1,101 @@
...
+  width: 11px;
+  height: 11px;
+  left: 13px;
+  top: 15px;
+  background-size: 11px;
+}
+
+:root {
+  --color: #3d4e61;
+}
```
VitaliyBondarenko1982, line 72: better to place :root selector on top of the file

VitaliyBondarenko1982 (approved):
Good job

src/style.css
hsvirina: Okay, I`ll replace

PR #3050, opened 2023-03-11
add task solution

LukashevychSergey:
- [DEMO LINK](https://Lukashevychsergey.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Lukashevychsergey.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,103 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: normal;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: bold;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: "Averin", sans-serif;
+}
+
+.search-form {
+  margin: 20px 0;
+}
+
+.image-search {
+  position: absolute;
+}
+
+.big {
+  width: 19px;
+  height: 19px;
+  margin-left: 26px;
+  margin-top: 25px;
+}
+
+.small {
+  width: 11px;
+  height: 11px;
+  margin-top: 15px;
+  margin-left: 13px;
+}
+
+.search-bar__field {
+  box-shadow: 0 1px 8px rgb(61 78 97 / 10%);
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  width: 100%;
+  height: 70px;
+  padding-left: 62px;
+  background: #fff;
+  outline: none;
+  font-size: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,103 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: normal;
```
Volodymyr-Melnyk-Fron10der, line 5: default values. you don't have to set those

src/style.css
```diff
@@ -0,0 +1,103 @@
...
+
+.small-search {
+  height: 42px;
+  padding-left: 33px;
+  font-size: 11px;
+  line-height: 19px;
+}
+
+.small-search::placeholder {
+  font-family: Avenir, sans-serif;
```
Volodymyr-Melnyk-Fron10der, line 98: just set it for html selector and it will be font by default

src/style.css
```diff
@@ -0,0 +1,103 @@
...
+  color: #3d4e61;
+}
+
+.search-bar__field:hover {
+  display: flex;
+  align-items: center;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px rgb(61 78 97 / 20%);
+  border-radius: 4px;
```
Volodymyr-Melnyk-Fron10der, line 65: you should change properties like colors, radius, etc. but not display or positioning or border (set only border-color)

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! keep it up! 

1. add text shadow as on design 
image
2. try to pass al tests

src/style.css
LukashevychSergey: Hi, by some strange reason its not working without putting it to all selectors because i already set it for all html, by the way i thought in this case i need to write font-family: inherit; to all selectors work the right way but its not working too(

src/style.css
LukashevychSergey: The tests in browser are all passed, what is the problem with git?

src/style.css
DarkMistyRoom: Remember that inputs and other interactive elements don’t inherit font styles by default. you should use `font-family: inherit;`

DarkMistyRoom (approved):
Good, just fix the font-family

src/style.css
LukashevychSergey: I already tried this but i can't pass the tests in browser because fonts stop working properly if i use 'inherit', and if i put font-family: Avenir; for all the selectors that i need its start working right.  And im really do not understand why...Maybe you got idea? (i checked everything out and found the mistake"i wrote Anevir" for html thats why inherit didn't work!)

PR #3052, opened 2023-03-11
solution

Kinderwagen:


src/index.html
```diff
@@ -11,6 +11,38 @@
...
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="https://www.google.com/"
+      method="get"
+      data-qa="big"
+      class="field"
+    >
+      <div class="searchBar">
```
etojeDenys, line 20: and let`s try to follow BEM or kebab-style class naming

etojeDenys (changes requested):
🔥

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  font-family: "Avenir";
+  font-style: normal;
+  font-weight: 900;
+  src: url(./fonts/Avenir-Heavy.ttf);
+}
+
+* {
+  box-sizing: border-box;
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
```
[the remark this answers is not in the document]
Kinderwagen: The attribute "style" attribute is cannot be used as it's banned - the problem is that i can't move these lines into html seletor

VitaliyBondarenko1982, line 18: this comment was to make like this
```suggestion
html {
    font-family: "Avenir", sans-serif;
   font-weight: 300;
}

* {
  box-sizing: border-box;

```

VitaliyBondarenko1982 (approved):

PR #3062, opened 2023-03-13
Develop

msukhovych:
DEMO https://msukhovych.github.io/layout_search-bar-airbnb/
TEST https://msukhovych.github.io/layout_search-bar-airbnb/report/html_report/


src/index.html
```diff
@@ -11,6 +11,47 @@
...
+          name="bigSearch"
+          placeholder="Try &quot;Los Angeles&quot;"
+          data-qa="keypress"
+          required
+        >
+        <img
+          class="big__icon icon"
+          src="./images/Search.svg"
+          alt="icon"
+        >
```
polosanya, line 33: Why do you put an icon after input? 
It would be more logical to put the icon first, like on the design. 
It will not confuse other developers and will reduce the needed indents for elements in CSS

src/style.css
```diff
@@ -0,0 +1,80 @@
+/* stylelint-disable color-hex-case */
```
polosanya, line 1: No need to disable the eslint. You can just fix the problem and convert color into the proper format

polosanya (changes requested):
Pay attention to the mockup. The text should be different on focus
image
image

polosanya (changes requested):
image
image
Also, input should take the whole width
image

tetiana-ivanchenko (approved):
placeholder still change styles on focus, if you have any question feel free to ask it in the corresponding chat in slack

PR #3065, opened 2023-03-14
Develop

MaksymKolesnyk:
https://MaksymKolesnyk.github.io/layout_search-bar-airbnb/
https://MaksymKolesnyk.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,8 +1,10 @@
+/*Add custom font*/
 @font-face {
   font-family: Avenir;
   src: url(./fonts/Avenir.ttc);
 }
 
+/* styles shered with the both search bar */
 .search-bar {
   display: flex;
   margin-top: 20px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,92 @@
...
+}
+
+.input-big-bar:focus {
+  outline: none;
+  background: linear-gradient(180deg,
+  rgb(255, 255, 255) 0%, rgb(246, 246, 247) 100%)
+   ;
+  font-weight: 600;
+  text-shadow: 0 4px 4px rgba(0, 0, 0, 0.25);
+}
```
anstsot, line 40: you have duplicated styles on line 80-87. it would be better to create a common class as `.input-bar` and add common styles fo that selector. then just use `.input-small-bar` to modify styles which are different for smaller input

anstsot (changes requested):
Good job, but there are few fixes to improve your code.

PR #3066, opened 2023-03-14
Develop

GlibVK:
Hi. Corrected all common class. 

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -18,7 +18,6 @@
       class="form"
       data-qa="big"
     >
-
       <img
         src="./images/Search.svg"
         alt="magnifying glass"
@@ -41,7 +40,6 @@
       class="form"
       data-qa="small"
     >
-
       <img
         src="./images/Search.svg"
         alt="magnifying glass"
--- a/src/index.html
+++ b/src/index.html
@@ -12,7 +12,7 @@
   </head>
 
   <body>
-    
+
     <form
       action="big"
       method="get"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -10,7 +10,50 @@
...
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="big"
+      method="get"
+      class="form"
+      data-qa="big"
+    >
+
+      <img
```
anstsot, line 22: don't add the empty line between parent and child
```suggestion
    >
      <img
```

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  top: 15px;
+  left: 13px;
+  bottom: 16px;
+  width: 11px;
+  height: 11px;
+}
+
+.input {
+  width: 100%;
+  font-family: Avenir-Heavy, sans-serif;
```
anstsot, line 48: please add the font-family style for whole `html` selector. here you can just use:
```suggestion
  font-family: inherit;
```

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  font-size: 16px;
+  line-height: 22px;
+  background: linear-gradient(180deg,#e5e8f0 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+}
+
+.input::placeholder {
+  font-family: Avenir, sans-serif;
+  font-style: normal;
```
anstsot, line 80: these styles aren't needed here. they should be inherited from `.input`

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  line-height: 22px;
+  font-weight: 400;
```
anstsot, line 82: don't add default styles (the `font-weight` is 400 by default)

anstsot (changes requested):
Good job, but there are few fixes to improve your code.

PR #3071, opened 2023-03-14
Add task solution

mariuszgit:
https://mariuszgit.github.io/layout_search-bar-airbnb/
https://mariuszgit.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,35 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+
+    <form
+      action="#"
+      method="POST"
+      class="form"
+      data-qa="big"
+    >
+      <input
+        name="searchInput"
+        class="form__input"
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+      >
+      <button type="submit" class="form__btn"></button>
+    </form>
+
+    <form
+      action="#"
+      method="POST"
+      class="form"
+      data-qa="small"
+    >
+      <input
+        name="searchInput"
+        class="form__input"
+        placeholder="Try &quot;Los Angeles&quot;"
+      >
+      <button type="submit" class="form__btn"></button>
+    </form>
+
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,73 @@
+* {
+  box-sizing: border-box;
+}
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: bold;
+  src: url(./fonts/Avenir-Heavy.ttf);
+}
+
+.form {
+  margin-top: 20px;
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,35 @@
...
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+      >
+      <button type="submit" class="form__btn"></button>
+    </form>
+
+    <form
+      action="#"
+      method="POST"
+      class="form"
```
DorotaLeniecDev, line 33: I think better sollution would be to style small element using modifier class.
You can add
```suggestion
      class="form form--small"
```
Instead of relying on **data-qa** for css selector which can change.

src/index.html
```diff
@@ -11,6 +11,35 @@
...
+      data-qa="small"
+    >
+      <input
+        name="searchInput"
+        class="form__input"
+        placeholder="Try &quot;Los Angeles&quot;"
+      >
+      <button type="submit" class="form__btn"></button>
```
DorotaLeniecDev, line 41: I see you are adding a search icon for submit button.
Usually search icons in this kind of forms are not submit buttons.
Please try to change this, and use different tag for adding search icon. You can try with **img** and **label**.

src/index.html
```diff
@@ -11,6 +11,35 @@
...
+      method="POST"
+      class="form"
+      data-qa="big"
+    >
+      <input
+        name="searchInput"
+        class="form__input"
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+      >
```
DorotaLeniecDev, line 26: Remember to add spaces between child elements :)

DorotaLeniecDev (changes requested):
Good PR, few fixes needed

src/style.css
```diff
@@ -57,17 +56,18 @@
...
   font-size: 14px;
   padding-left: 33px;
 }
 
-.form[data-qa="small"] .form__btn {
-  width: 11px;
-  height: 11px;
-  background-size: 11px 11px;
-  background-repeat: no-repeat;
+.form--small .form__btn {
```
DorotaLeniecDev, line 65: Sorry I have missed that. According to checklist you should try your best not to increase css selector specificity,
So instead of **.form--small .form__btn** you can add new class to style thise elements :)

src/style.css
```diff
@@ -57,17 +56,18 @@
...
   left: 13px;
   top: 15px;
 }
+
+.form--small .form__image {
```
DorotaLeniecDev, line 70: Also same here

src/style.css
```diff
@@ -57,17 +56,18 @@
   outline: none;
 }
 
-.form[data-qa="small"] > .form__input {
+.form--small .form__input {
```
DorotaLeniecDev, line 59: And here

DorotaLeniecDev (changes requested):
Sorry I have missed one thing, see my comments and make changes and we are good with this PR :)

Radoslaw-Czerniawski (approved):
Some comments from my side for the future, pr looks and works like it should, good job :D

PR #3072, opened 2023-03-14
Fix2

VadymKapeliusnyi:


src/style.css
```diff
@@ -0,0 +1,59 @@
...
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
+
+* {
+  box-sizing: border-box;
+  outline: none;
+  font-family: "Avenir", sans-serif;
+  color: rgba(61, 78, 97, 1);
```
vlad-demchuk, line 19: move outline and color to special elements you need to change
don't apply it for all elements (`*`)

src/style.css
```diff
@@ -0,0 +1,59 @@
...
+  background-repeat: no-repeat;
+  width: 100%;
+  border: 1px solid rgba(225, 231, 237, 1);
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+  transition: all 0.3s;
+  background-position: 26px 25px, 0 0;
+  text-indent: 62px;
+  height: 70px;
+  font-size: 16px;
```
vlad-demchuk, line 38: don't forget about css properties order
http://codeguide.academy/html-css.html#css-order

vlad-demchuk (changes requested):
fix and add demo and test report links to PR description

PR #3074, opened 2023-03-14
Develop

GlibVK:


polosanya (changes requested):
Almost done!
The font should be bold only on focus. Pay attention to the design
image
image

PR #3075, opened 2023-03-14

ArtemBehter:


src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 900;
+}
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+html {
```
DorotaLeniecDev, line 16: I think that better practice is to set global **font-family** on **body** element as all content lies under body. You can check this discussion what is best practice in this post: https://stackoverflow.com/questions/42925682/where-to-set-font-family-body-or-html-element

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+}
+
+.search-icon-small {
+  border: none;
+  display: block;
+  position: absolute;
+  background-color: transparent;
+  left: 10px;
+  top: 10px;
+}
```
DorotaLeniecDev, line 41: You can create default styles for **search-icon** and add modifier like **search-icon-small** and add there styles that applies only for small icon.

```suggestion
.search-icon {
  border: none;
  display: block;
  position: absolute;
  left: 26px;
  top: 25px;
  background-color: transparent;
}

.search-icon-small {
  left: 10px;
  top: 10px;
}
```

- [ ] combine classes for common layput
- [ ] adjust search-icon-small position to be aligned with the designs
- [ ] add proper class names in HTML - element can have more than one class

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  line-height: 19px;
+  color: rgb(61, 78, 97);
+  text-shadow: none;
+}
+
+.small_input_search {
+  font-size: 14px;
+  line-height: 19px;
+  height: 42px;
+  box-sizing: border-box;
```
DorotaLeniecDev, line 96: You are setting **box-sizing: border-box;** to all emenents already using * selector (top of this document) so there is no need to repeat it here agian

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  padding-left: 33px;
+  background-size: 11px;
+  background-position: 13px 15px;
+}
+
+.big_input_search:hover,
+.small_input_search:hover {
```
DorotaLeniecDev, line 103: You can use common class here to style big and small input at once **search-input**

src/style.css
```diff
@@ -0,0 +1,114 @@
...
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px  rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.big_input_search:focus,
+.small_input_search:focus {
```
DorotaLeniecDev, line 110: Same here

DorotaLeniecDev (changes requested):
Good job but still some changes needed. Please check my comments

src/index.html
```diff
@@ -11,6 +11,48 @@
...
-    <h1>Search bar airbnb</h1>
+
+    <form
+      class="form-search"
+      action="#"
+      method="get"
+      data-qa="big"
+    >
+
+      <button type="submit" class="search-icon-big">
```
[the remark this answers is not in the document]
ArtemBehter: How can I do it ?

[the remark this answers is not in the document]
DorotaLeniecDev: For example:

`        <label for="search" class="search-form__label">
            Search icon
        </label>`

src/style.css
```diff
@@ -0,0 +1,110 @@
...
+
+.search-icon {
+  border: none;
+  display: block;
+  position: absolute;
+  background-color: transparent;
+}
+
+.search-icon-big {
+  left: 26px;
```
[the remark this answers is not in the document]
Radoslaw-Czerniawski: Same for the smaller one, however the calc will be different

Radoslaw-Czerniawski (changes requested):
Needs adjustments, look how you should structure indents according to the checklist

src/style.css
```diff
@@ -0,0 +1,112 @@
...
+}
+
+.big_input_search {
+  height: 70px;
+  line-height: 22px;
+  font-size: 16px;
+  box-sizing: border-box;
+  background-repeat: no-repeat;
+  background-size: 19px;
+  background-position: 26px 25px;
```
DorotaLeniecDev, line 70: Do you still need those? If no then remove

src/style.css
```diff
@@ -0,0 +1,112 @@
...
+  padding: 25px 0 26px 62px;
+  margin: 20px 0;
```
DorotaLeniecDev, line 72: I think you are setting margin already for form-search class so this is not needed here

src/style.css
```diff
@@ -0,0 +1,112 @@
...
+}
+
+.small_input_search::placeholder {
+  font-family: inherit;
+  font-weight: 300;
+  font-size: 14px;
+  line-height: 19px;
+  color: rgb(61, 78, 97);
+  text-shadow: none;
+}
```
DorotaLeniecDev, line 91: Try to combine common styles for both placeholders
```suggestion

.big_input_search::placeholder,
.small_input_search::placeholder {
  font-family: inherit;
  font-weight: 300;
  color: rgb(61, 78, 97);
  text-shadow: none;
}
```

src/style.css
```diff
@@ -0,0 +1,112 @@
...
+
+.small_input_search {
+  font-size: 14px;
+  line-height: 19px;
+  height: 42px;
+  padding-left: 33px;
+  background-size: 11px;
```
DorotaLeniecDev, line 98: It this needed? If not then remove unused styles

src/style.css
```diff
@@ -0,0 +1,112 @@
...
+  background-position: 13px 15px;
```
DorotaLeniecDev, line 99: It this needed? If not then remove unused styles

DorotaLeniecDev (approved):

PR #3079, opened 2023-03-15
solution

VadymKapeliusnyi:
https://vadymkapeliusnyi.github.io/layout_search-bar-airbnb/
https://VadymKapeliusnyi.github.io/layout_search-bar-airbnb/report/html_report/


src/style.css
```diff
@@ -0,0 +1,59 @@
...
+
+.search-input {
+  height: 70px;
+  width: 100%;
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+  text-indent: 62px;
+  color: rgba(61, 78, 97, 1);
+  background-position: 26px 25px, 0 0;
+  background-image: url(images/Search.svg);
```
slavanosatskyi, line 31: just add html element img with link to this icon. Then you won't need to copy it here and in search-input:focus

src/style.css
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.search-form {
+  margin-top: 20px;
+}
+
+.search-input {
+  height: 70px;
+  width: 100%;
+  font-family: "Avenir", sans-serif;
```
slavanosatskyi, line 26: usually it is defined in html selector

slavanosatskyi (changes requested):
Nice work, but try to create loupe icon as html element

src/style.css
VadymKapeliusnyi: image

yes but I should then skip the default values ​​of the browser in *

VadymKapeliusnyi: > usually it is defined in html selector


maybe i don't understand something

PR #3080, opened 2023-03-15
create HTML page with two search bars

pos1bl:
- [DEMO LINK](https://pos1bl.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://pos1bl.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,102 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir.ttc);
+  font-style: normal;
+}
+
+* {
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  --big-height: 70px;
+  --small-height: 42px;
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
+
+.input {
+  background: #fff;
+  height: 100%;
+  width: 100%;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,102 @@
...
+  padding-left: 33px;
+  line-height: 19px;
+  outline: none;
+  font-family: inherit;
+  font-weight: 300;
+  color: #3d4e61;
+  font-size: 14px;
+}
+
+.inputBig {
```
slavanosatskyi, line 35: Try to rework it according to checklist. Just create one class for input and then specification in input--small 
Screenshot 2023-03-15 at 17 28 48

slavanosatskyi (changes requested):

src/style.css
```diff
@@ -0,0 +1,110 @@
...
+.label::after {
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  position: absolute;
+  content: "";
+  display: block;
+}
+
+.label--big::after {
```
vlad-demchuk, line 59: the same, don't use `--big` and `--small`

vlad-demchuk: check all such cases

vlad-demchuk (changes requested):

nazarmatsevych (changes requested):
Almost done,
pls check comment

PR #3081, opened 2023-03-15
 layout search bar 

Aleksandra-hub:
I have two problems : with small search bar text, it did pass test, and focus statement , logo of search did appear above background color.


IvanFesenko (changes requested):
First. You should create develop branch and make changes there.
Demo link isn't working
If you have any questions - you should ask them in fe_chat

etojeDenys (changes requested):
i don't see any changes

PR #3107, opened 2023-03-20
add search bar airbnb

Leva313:
Were is my problem?


VitaliyBondarenko1982 (approved):
good job.
about tests - something went wrong with reference image.
if somithing like that will be in next tasks - try to remove backstop_data folder and run npm test again

Leva313:
Thanks for the advice)

PR #3112, opened 2023-03-20
search bar airbnb

shelyaa:
https://shelyaa.github.io/layout_search-bar-airbnb/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,12 +1,15 @@
-/* stylelint-disable font-family-no-missing-generic-family-keyword */
 @font-face {
-  font-family: Avenir-Book;
+  font-family: "Avenir-Book";
   src: url(fonts\Avenir-Book.ttf);
+  font-weight: 300;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,120 @@
...
+  height: 19px;
+  margin: 25px 17px 26px 26px;
+}
+
+.img-2 {
+  width: 11px;
+  height: 11px;
+  margin: 15px 9px 16px 13px;
+}
+
```
tetiana-ivanchenko, line 52: change classnames according to checklist
also, dont use tags as css selectors

tetiana-ivanchenko (changes requested):

src/style.css
tetiana-ivanchenko: also placeholders shouldn't change styles on focus

VitaliyBondarenko1982 (approved):

PR #3119, opened 2023-03-21
Solution

yao157:


src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  color: #3d4e61;
+  font: inherit;
+}
+
+.input:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.input:focus {
+  font: inherit;
```
YegorKorenkov, line 60: Repeated, you already add this rule for .input selector

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.search-image-small:focus {
+  font-size: 14px;
+  background: url(./images/Search.svg) no-repeat 13px 50%/11px,
+    linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
+
+.input:focus::placeholder {
+  font-weight: 300;
```
YegorKorenkov, line 77: font-weight alredy 300. Please try to add effect wich will be visible

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+.search-image-small {
+  height: 42px;
+  padding: 0 33px;
+  font-size: 14px;
+  background: url(images\Search.svg) no-repeat 13px 50%/11px;
+}
+
+.input::placeholder {
+  color: #3d4e61;
+  font: inherit;
```
YegorKorenkov, line 52: Same

PR #3132, opened 2023-03-28
add task one

obadyda:
https://obadyda.github.io/layout_search-bar-airbnb/
https://obadyda.github.io/layout_html-form/report/html_report/


src/style.css
```diff
@@ -0,0 +1,100 @@
...
+  margin: 20px auto 0;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  box-sizing: border-box;
+  position: relative;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  padding: 0 0 0 62px;
+  font-size: 16px;
+  line-height: 22px;
+  font-family: "Avenir", sans-serif;
```
etojeDenys, line 35: move to html selector

src/style.css
```diff
@@ -0,0 +1,100 @@
...
+
+.equip-bar {
+  display: flex;
+  justify-content: center;
+  width: 100%;
+  height: 70px;
+  margin: 20px auto 0;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  box-sizing: border-box;
```
etojeDenys, line 29: move to `*` selector

etojeDenys (changes requested):
let's create a form class that will have styles from two forms to make our code easier to read and refactor in the future
(we can also do this for `label` & `input`)

src/style.css
obadyda: > move to html selector

when i do this the tests fail

etojeDenys (approved):
good job

PR #3135, opened 2023-03-29
add task solution

Platon-jun:
DEMO LINK](https://platon-jun.github.io/layout_search-bar-airbnb/)
 [TEST REPORT LINK](https://platon-jun.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,123 @@
+@font-face {
+  font-family: Avenir;
+  font-weight: 300px;
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: bold;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+.body {
+  margin-top: 0;
+}
+
+.container {
+  margin-top: 20px;
+  position: relative;
+}
+
+.search-field.big {
+  width: 100%;
+  height: 70px;
+  background-color: #fff ;
+  border-radius: 4px;
+  border: solid 1px #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  font-size: 16px;
+  font-family: Avenir, sans-serif;
+  font-weight: 300px;
+  padding-left: 62px;
+}
+
+.search-field.small {
+  width: 100%;
+  height: 42px;
+  background-color: #fff ;
+  border-radius: 4px;
+  border: solid 1px #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  font-family: Avenir, sans-serif;
+  font-weight: 300px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,123 @@
...
+  border-radius: 4px;
+  border: solid 1px #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  font-family: Avenir, sans-serif;
+  font-weight: 300px;
+  font-size: 14px;
+  padding-left: 33px;
+}
+
+.search-field.big::placeholder {
```
vlad-demchuk, line 48: do not increase css specificity
use just one class instead of `search-field.big`

vlad-demchuk: fix all such cases

vlad-demchuk (changes requested):
GJ, just small improvements

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  position: relative;
+}
+
+.search-field {
+  width: 100%;
+  height: 70px;
+  background-color: #fff ;
+  border-radius: 4px;
+  border: solid 1px #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
```
OleziO, line 28: Try to use HEX instead of RGBA

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  font-size: 16px;
+  font-family: Avenir, sans-serif;
```
OleziO, line 30: Move it to `html` selector and remove it from all other places.

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+
+.search-field.small {
+  height: 42px;
+  font-size: 14px;
+  padding-left: 33px;
+}
+
+.search-field::placeholder {
+  font-weight: 300px;
+  color: #3d4e61;
```
OleziO, line 43: Move all repeatable colors to CSS variables.

OleziO (changes requested):

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+}
+
+.search-field {
+  color: var(--main-text-color);
+  width: 100%;
+  height: 70px;
+  border-radius: 4px;
+  border: solid 1px #e1e7ed;
+   font-size: 16px;
+  font-family: Avenir, sans-serif;
```
vlad-demchuk, line 37: move `font-family` to `html`
we usually store common font styles in `html`
```
html {
  font-family: Avenir, sans-serif;
}

vlad-demchuk (changes requested):

src/style.css
Platon-jun: when i move font-family to html, it`s not wurking.

IvanFesenko (approved):

PR #3139, opened 2023-03-30
create a serach bar

yevsss:
[DEMO LINK](https://yevsss.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://yevsss.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,43 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      data-qa="big"
+      action="#"
+      method="post"
+      >
+      <label class="searchBar-1">
+        <img
+          class="searchIcon-1"
+          src="images/Search.svg"
+          alt="image"
+        >
+        <input
+          class="search-1"
+          type="search"
+          data-qa="keypress"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
+
+    <form
+      action="#"
+      method="post"
+      data-qa="small"
+    >
+      <label class="searchBar-2">
+        <img
+          class="searchIcon-2"
+          src="images/Search.svg"
+          alt="image"
+        >
+        <input
+          class="search-2"
+          type="search"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,127 @@
+@font-face {
+  font-family: "Avenir";
+  font-style: normal;
+  font-weight: 300;
+  src: url("fonts/Avenir-Book.ttf");
+}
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,43 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      data-qa="big"
+      action="#"
+      method="post"
+      >
```
OleziO, line 18: Remove two spaces before `>`

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+}
+
+.search-1 {
+  width: 100%;
+  position: relative;
+  padding-left: 62px;
+  height: 70px;
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
```
OleziO, line 34: Try to use HES instead of RGBA

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  display: block;
+}
+
+.search-1 {
+  width: 100%;
+  position: relative;
+  padding-left: 62px;
+  height: 70px;
+  background: #fff;
+  border: 1px solid #e1e7ed;
```
OleziO, line 33: Move all repeatable colors to CSS variables. CSS variables must be in the `:root` selector.

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  color: #3d4e61;
+  text-shadow: 0 4px 4px rgba(0, 0, 0, 0.25);
+}
+
+.searchBar-2 {
+  height: 42px;
+  position: relative;
+}
+
+.search-2 {
```
OleziO, line 72: Rename search-2 and other classes to more meaningful names.

OleziO (changes requested):

src/index.html
yevsss: Changed

src/style.css
yevsss: Maybe you meant HEX?
Don't think that it's a good idea, because:
1. The designer gave us color in RGBA format
2. If You do not agree, tell me how to translate colors RGBA in HEX 
3. Tried this one: https://simplecss.eu/rgbatohex.html =>working not good

Thank you!)

src/style.css
yevsss: I did it, HOWEVER, I don't think that moving color that repeats twice is a good idea, for me it is uncomfortable + it makes the length of the code line more than 80 symbols

src/style.css
yevsss: UPD: Did it, but it is the designer`s responsibility

src/index.html
yevsss: Removed

etojeDenys (approved):
good job

PR #3144, opened 2023-03-31
First

ColumbusDeveloper:


YegorKorenkov (changes requested):
1. Please provide classes for all elements
2. Make small input fullwidth
image
3. 
image
4. Please also make different hover, focus effects
5. Pleas read again requirements, checklist and fix all

PR #3151, opened 2023-04-01
let me cook 

timohahaaa:
https://timohahaaa.github.io/layout_search-bar-airbnb/

https://timohahaaa.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -7,7 +7,7 @@ html {
 }
 
 body {
-  font-family: "Avenir", sans-serif;
+  font-family: inherit;
   font-size: 16px;
 }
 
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,104 @@
...
+}
+
+.iconSmall {
+  position: absolute;
+  left: 13px;
+  top: 15px;
+  display: block;
+  width: 11px;
+  height: 11px;
+}
```
polosanya, line 104: image

polosanya (changes requested):
Good job! Let's try to improve your code a bit)
Also, it would be great to pass all tests
image
image
Font doesn't match the design
image
image

PR #3155, opened 2023-04-02
add task

Duster-1:
https://duster-1.github.io/layout_search-bar-airbnb/
https://duster-1.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,118 @@
+* {
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir-bold";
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,118 @@
...
+}
+
+.input--small:focus {
+  border-radius: 3px;
+  color: #3d4e61;
+  font-style: normal;
+  font-weight: 800;
+  font-size: 14px;
+  line-height: 19px;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
OleziO, line 117: Move all repeatable colors to CSS variables and move them to :root selector.

[More info. ](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

OleziO (changes requested):
Demo: 
image
Design: 
image

nazarmatsevych (approved):
Good, but remove outline pls, you can use outline: none rule
image

PR #3170, opened 2023-04-04
Develop

VasylKurtiak:


readme.md
```diff
@@ -1,20 +1,24 @@
 # Search bar for Airbnb
+
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_search-bar-airbnb/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_search-bar-airbnb/report/html_report/)
+
+- [DEMO LINK](https://VasylKurtiak.github.io/layout_search-bar-airbnb/)
+- [TEST REPORT LINK](https://VasylKurtiak.github.io/layout_search-bar-airbnb/report/html_report/)
```
Viktor-Kost, line 6: please add links to Pull Request description

src/index.html
```diff
@@ -10,7 +10,33 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <form action="#" method="post" class="form-field">
```
Viktor-Kost, line 15: there is should be two different forms instead of one
```
<form>
  <label>
    <img>
    <input>
  </label>
</form>

<form>
  <label>
    <img>
    <input>
  </label>
</form>
```

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 900;
+}
+
+.input {
+  width: 100%;
+  height: 70px;
+  margin-top: 10px;
+  margin-bottom: 20px;
```
Viktor-Kost, line 23: apply `margin-bottom` to `.form` instead of `.input`

You should avoid margin collapsing, when a child margins go through a parent and push off another parent
[useful_link](https://www.smashingmagazine.com/2019/07/margins-in-css/#parent-and-first-or-last-child-element)
So, if you want one element to be push off from another element use margins at THESE elements, not their children

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+}
+
+.inputDefaultSmall:hover {
+  background: ffffff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
```
Viktor-Kost, line 63: apply to hover/focus states only those rules that are different from static one

Viktor-Kost (changes requested):
Hello!)

You have a lot of repeatable styles. 

Following the BEM-methodology - use the `block` selector for the FIRST input (`class="form__input"`), where you apply common form styles

```
.form_input {
 {{styles_that_are_common_for_both_inputs}}
}
```

And then you apply `modifier` to the SECOND form (`class="form__input form__input--small"`) that have ONLY differences with base class form
```
.form_input--small {
 {{styles_that_are_different_from_form__input}}
}
```
So in this way you follow DRY (Don't repeat yourself) principle

src/style.css
```diff
@@ -0,0 +1,87 @@
+body {
```
Viktor-Kost, line 1: remove default margin from body

src/index.html
```diff
@@ -10,7 +10,33 @@
...
+        <input 
+          data-qa="big"
+          class="input"
+          type="text" 
+          name="place"
+          placeholder="Try &quot;Los Angeles&quot;"
+          required  
+        >
+      </label>
+      <label class="smallInput">
```
Viktor-Kost, line 29: add a loop to the second input as well
Screenshot 2023-04-04 at 20 55 53

Viktor-Kost (changes requested):

readme.md
VasylKurtiak: Thanks for the tips) I will fix it now.

src/index.html
VasylKurtiak: What did you mean by this task?

PR #3175, opened 2023-04-05
add task solution

SerhiiVizir:
- [DEMO LINK](https://SerhiiVizir.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://SerhiiVizir.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,65 @@
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: normal;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: bold;
+}
+
+html {
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+}
+
+.input {
+  display: flex;
+  background-image: url(images/Search.svg);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,69 @@
...
+  font-weight: bold;
+}
+
+.input::placeholder {
+  font-family: Avenir, sans-serif;
+  color: #3d4e61;
+}
+
+.input:focus::placeholder {
+  color: transparent;
```
vlad-demchuk, line 47: why do you hide placeholder? it fails a test

vlad-demchuk (approved):

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+
+.searchbar-big {
+  height: 70px;
+  font-size: 16px;
+  padding-left: 62px;
+  background-position: 26px 25px;
+  line-height: 22px;
+}
+
+.searchbar-small {
```
vlad-demchuk, line 62: check this checklist point
image

vlad-demchuk (changes requested):

src/style.css
SerhiiVizir: this test didn`t pass before I wrote this part of code

IvanFesenko (approved):

PR #3197, opened 2023-04-07
add task solution

maksym-nemera:
https://maksym-nemera.github.io/layout_search-bar-airbnb/
https://maksym-nemera.github.io/layout_search-bar-airbnb/report/html_report/


naziamoff (changes requested):
Not bad, but there are important suggestions
Except for those, your fonts don't seem to change when the input is focused, but they do accordingly to the mockup

naziamoff (approved):
🔥

PR #3210, opened 2023-04-09
Develop5

ColumbusDeveloper:


src/style.css
```diff
@@ -0,0 +1,57 @@
...
+  box-shadow: none;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  border: 1px solid #e1e7ed;
+  border-radius: 3px;
+  text-shadow: 0 4px 4px rgba(0, 0, 0, 0.2);
+}
+
+.input-small {
+  font-size: 14px;
+  line-height: 19.12px;
```
nazarmatsevych, line 49: Don't apply decimal values, no need to copy all styles from mockup

nazarmatsevych (changes requested):
check comments and proofread your code with [checklist](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md)

Also image should not disappear on focus
image

src/style.css
vlad-demchuk: https://github.com/mate-academy/layout_search-bar-airbnb/pull/3210/files#r1161302562

vlad-demchuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,60 @@
...
+@font-face {
+  font-family: "Avenir";
+  src: url("./fonts/AvenirHeavy.woff") format("woff");
+  font-style: normal;
+  font-weight: 800;
+}
+
+.input {
+  width: 100%;
+  font-family: Avenir, Arial, Helvetica, sans-serif;
```
YegorKorenkov, line 17: it would be better to move it to html selector

YegorKorenkov (approved):
Looks good

PR #3228, opened 2023-04-13
add task solution

Krivets:
2 тест не проходить, не можу зрозуміти в чому проблема

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,43 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="qw"
+      method="post"
+      class="conteiner-1"
+      data-qa="big"
+      >
+      <img
+        src="images/Search.svg"
+        alt="images"
+        class="svg-img-1"
+        >
+      <input
+        type="text"
+        name="bar1"
+        class="bar-1"
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+      >
+    </form>
+
+    <form
+      action="qw"
+      method="post"
+      class="conteiner-2"
+      data-qa="small"
+      >
+      <img
+        src="images/Search.svg"
+        alt="images"
+        class="svg-img-2"
+        >
+      <input
+        type="text"
+        name="bar2"
+        class="bar-2"
+        placeholder="Try &quot;Los Angeles&quot;"
+      >
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,125 @@
+@font-face {
+  font-family: Avenir-book;
+  src: url(./fonts/Avenir-Book.ttf) format("truetype");
+  font-weight: lighter;
+}
+@font-face {
+  font-family: Avenir-Heavy;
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,125 @@
...
+  width: 100%;
+  padding: 0;
+  border: none;
+  outline: none;
+  font-family: "Avenir", sans-serif;
+  font-style: normal;
+  font-weight: 400;
+  font-size: 14px;
+  line-height: 19px;
+}
```
anstsot, line 94: you have some similar styles in `.bar-1` selector. you can create a common class `.input` and merge the same styles.

anstsot: same for other elements (`form` and `img`)

src/style.css
```diff
@@ -0,0 +1,125 @@
...
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir.ttc) format("truetype");
+  font-weight: normal;
+}
+
+body {
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
```
anstsot, line 19: you haven't included font "Avenir" with weight 300 that's why it's not necessary.

anstsot (changes requested):
It’s almost done! Let’s improve a little.
Please add proper styles for the focus effect
image

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+  font-weight: bold;
+}
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir.ttc) format("truetype");
+  font-weight: 300;
+}
+
+body {
+  font-family: "Avenir", sans-serif;
```
witflash, line 18: You can add `font-weight: 300;` to use by default for the whole document and `font-weight: inherit` for the `input` elements

src/style.css
```diff
@@ -0,0 +1,96 @@
+@font-face {
+  font-family: Avenir-book;
+  src: url(./fonts/Avenir-Book.ttf) format("truetype");
+  font-weight: lighter;
+}
```
witflash, line 5: You don't use this font as I see. So you can remove it at all

src/style.css
```diff
@@ -0,0 +1,96 @@
...
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+  border-radius: 4px;
+}
+
+.input {
+  width: 100%;
+  border: none;
+  outline: none;
+  padding: 0;
+  font-family: "Avenir", sans-serif;
```
witflash, line 54: Add here `font-weight: inherit;` to use font-weight from the body

witflash (changes requested):
Good job!
Check my comments, please. It may help you to fix the last test.

src/index.html
```diff
@@ -5,12 +5,49 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="qw"
+      method="post"
+      class="conteiner-1 form"
+      data-qa="big"
+      >
```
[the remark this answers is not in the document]
Krivets: після виправлень залишилась та сама помилка

PR #3243, opened 2023-04-17
add two search

dimarogkov:
https://dimarogkov.github.io/layout_search-bar-airbnb
https://dimarogkov.github.io/layout_search-bar-airbnb/report/html_report


src/style.css
```diff
@@ -0,0 +1,117 @@
...
+  font-display: swap;
+}
+
+body,
+body * {
+  font-family: 'Avenir', sans-serif;
+  padding: 0;
+  margin: 0;
+  box-sizing: border-box;
+}
```
anstsot: do not set the font-family, margin and padding to all elements on the page. choose only that you need

anstsot (approved):
Good job, approving. But please keep attention to one comment I've left

src/style.css
dimarogkov: Done. Fix it at new commit

PR #3247, opened 2023-04-18
add task solution

KatyaZaremska:
- [DEMO LINK](https://KatyaZaremska.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://KatyaZaremska.github.io/layout_search-bar-airbnb/report/html_report/)


naziamoff (changes requested):
Good job in general 👍 

Seems the font does not change when I'm entering the input. But accordingly to the mockup - it should

PR #3264, opened 2023-04-19
task solution

julshvets:
- [DEMO LINK](https://julshvets.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://julshvets.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,48 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="http://localhost:3000/api"
+      method="post"
+      class="searchform"
+      data-qa="big"
+    >
+      <label class="searchbar">
+        <img
+          src="images/Search.svg"
+          alt="Search_icon"
+          class="icon icon-big"
+        >
+        <input
+          type="search"
+          name="Search"
+          data-qa="keypress"
+          placeholder="Try &quot;Los Angeles&quot;"
+          class="input input-big"
+          >
+      </label>
+    </form>
+
+    <form
+      action="http://localhost:3000/api"
+      method="post"
+      class="searchform"
+      data-qa="small"
+    >
+      <label class="searchbar">
+        <img
+          src="images/Search.svg"
+          alt="Search_icon"
+          class="icon icon-small"
+        >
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,48 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="http://localhost:3000/api"
+      method="post"
+      class="searchform"
```
naziamoff, line 17: use either `camelCase` or `kebab-case` for your classnames. It helps to keep your code consistent and easy to read

src/index.html
```diff
@@ -11,6 +11,48 @@
...
+          alt="Search_icon"
+          class="icon icon-big"
+        >
+        <input
+          type="search"
+          name="Search"
+          data-qa="keypress"
+          placeholder="Try &quot;Los Angeles&quot;"
+          class="input input-big"
+          >
```
naziamoff, line 32: closing brace should be at the same level as opening one - this way your layout is better structured and it is easier to distinguish where ends one tag and the other starts
```suggestion
        >
```

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+  left: 26px;
+  top: 25px;
+}
+
+.icon-small {
+  width: 11px;
+  height: 11px;
+  left: 13px;
+  top: 15px;
+}
```
naziamoff, line 82: see [checklist](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md) p.2

naziamoff (approved):
Good job 👍 
There are some suggestions though, so make sure to keep them in mind in your following Pull Requests

- placeholder should not be changed on focus, it should stay default. Only text you enter should be different

src/index.html
julshvets: Fixed.

src/index.html
julshvets: Fixed.

src/style.css
julshvets: There should be no class "icon-big", just "icon" and "icon-small". 
Fixed.
Thank you for checking my task.

PR #3275, opened 2023-04-21
add task solution

sasdaiv:
[DEMO LINK](https://sasdaiv.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://sasdaiv.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,87 @@
+* {
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 900;
+}
+
+.form {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
...
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 900;
+}
+
+.form {
+  font-family: "Avenir", sans-serif;
```
DarkMistyRoom, line 18: set font-family once to the `html `tag and inherit where needed

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+.search__small::placeholder {
+  font-size: 14px;
+}
+
+.form__small {
+  height: 42px;
+}
+
+.img-small {
+  position: absolute;
```
DarkMistyRoom, line 82: no need to duplicate the position. specify only those styles which are different

DarkMistyRoom (approved):
Well done

PR #3280, opened 2023-04-24
Develop

Exess94:
https://Exess94.github.io/layout_search-bar-airbnb/
https://Exess94.github.io/layout_search-bar-airbnb/report/html_report/


src/index.html
```diff
@@ -11,6 +11,55 @@
...
+    >
+      <label class="big-search">
+        <div
+          class="search-bar-big"
+          data-qa="big"
+        >
+          <img src="./images/Search.svg"
+            alt="123"
+            class="big-image"
+          >
```
sTorba24, line 26:
```suggestion
          <img 
            src="./images/Search.svg"
            alt="123"
            class="big-image"
          >
```

src/index.html
```diff
@@ -11,6 +11,55 @@
...
+    >
+      <label class="label-small-search">
+        <div
+          class="search-bar-small"
+          data-qa="small"
+        >
+          <img src="./images/Search.svg"
+            alt="123"
+            class="small-image"
+          >
```
sTorba24, line 50:
```suggestion
          <img 
            src="./images/Search.svg"
            alt="123"
            class="small-image"
          >
```

src/style.css
```diff
@@ -0,0 +1,142 @@
...
+}
+
+body {
+  margin: 0;
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
+
+input {
+  box-sizing: content-box;
```
sTorba24, line 23: set it for *

src/style.css
```diff
@@ -0,0 +1,142 @@
...
+}
+
+.big-search-bar-input {
+  position: relative;
+  left: 16px;
+  font-size: 16px;
+  font-weight: 400;
+  width: calc(100% - 50px);
+  color: #3d4e61;
+  font-family: "Avenir", sans-serif;
```
sTorba24, line 54: use `inherit` value here
Fix everywhere

sTorba24 (changes requested):
Good work! lets improve it

src/index.html
```diff
@@ -11,6 +11,57 @@
...
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+     action="http://localhost:3000/api"
+     method="post"
+    >
+      <label class="big-search">
+        <div
+          class="search-bar-big"
```
nazarmatsevych, line 20: According to this checklist part you can create common class `search-bar` and move all dublicate styles under it.
Add this class to both forms and modify styles only for small form using class-modifier `search-bar--small`

image

Fix all such cases

nazarmatsevych (changes requested):

src/style.css
```diff
@@ -0,0 +1,142 @@
...
+  font-weight: 800;
+  font-size: 16px;
+  line-height: 22px;
+  color: #3d4e61;
+  outline: none;
+  border: none;
+}
+
+.search-bar--small {
+  position: relative;
```
vlad-demchuk, line 78: no need to repeat all styles from `.search-bar`

apply only those that differ
check the previous comment

vlad-demchuk (changes requested):

src/style.css
vlad-demchuk: fix all such cases

src/style.css
```diff
@@ -0,0 +1,142 @@
+/* stylelint-disable shorthand-property-no-redundant-values */
```
IvanFesenko, line 1: Fix error instead of disable lint

src/style.css
```diff
@@ -0,0 +1,142 @@
...
+  height: 100%;
+  width: calc(100% - 60px);
+  font-size: 16px;
+  font-weight: 300;
+  left: 16px;
+  top: -1px;
+  outline: none;
+  border: none;
+  padding: 0;
+  font-family: "Avenir", sans-serif;
```
IvanFesenko, line 36:
```suggestion
  font-family: inherit;
```

src/style.css
IvanFesenko: Not fixed

src/style.css
```diff
@@ -0,0 +1,142 @@
...
+  max-height: 70px;
+  line-height: 100%;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+}
+
+.input::placeholder {
+  font-size: 16px;
+  line-height: 22px;
+  color: #3d4e61;
```
IvanFesenko, line 60: Move repeatable colors into variables

src/style.css
```diff
@@ -0,0 +1,142 @@
...
+  left: 18px;
+  height: 36px;
+  line-height: 19px;
+  font-weight: 400;
+  font-size: 14px;
+  font-family: "Avenir", sans-serif;
+  width: calc(100% - 36px);
+  outline: none;
+  border: none;
+}
```
IvanFesenko, line 105: Same here. Do not repeat your self. 

2023-04-27_15h51_41

IvanFesenko (changes requested):
Refactor required

src/style.css
OleziO: Move all repeatable colors to CSS variables.

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+
+.input--small::placeholder {
+  font-size: 14px;
+  line-height: 19px;
+  font-weight: 400;
+  color: #3d4e61;
+}
+
+.search-bar--small:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.1);
```
OleziO, line 102: Try to use HEX instead of RGBA

OleziO (changes requested):

src/index.html
```diff
@@ -11,6 +11,57 @@
...
+          >
+          <input
+            class="input"
+            name="default-search-bar"
+            type="text"
+            placeholder="Try &quot;Los Angeles&quot;"
+          >
+        </div>
+
+      </label>
```
polosanya, line 36:
```suggestion
        </div>
      </label>
```
Don't leave empty lines between parent and child elements. Also, IMHO, div is redundant here, as you already have label as a container

src/index.html
```diff
@@ -11,6 +11,57 @@
...
+
+    </form>
+
+
   </body>
```
polosanya, line 66:
```suggestion
        </div>
      </label>
    </form>
  </body>
```

polosanya (changes requested):
Almost perfect!
image
image

VitaliyBondarenko1982 (approved):

PR #3287, opened 2023-04-25
add task solution

bardyuzha:
https://bardyuzha.github.io/layout_search-bar-airbnb/
https://bardyuzha.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -16,7 +16,10 @@
 
 * {
   box-sizing: border-box;
-  outline: none;
+}
+
+html {
+  font-family: Avenir, sans-serif;
 }
 
 body {
@@ -24,12 +27,12 @@ body {
   font-weight: 300;
 }
 
-.search {
+.form {
   position: relative;
   margin-bottom: 20px;
 }
 
-.search object {
+.form .search_icon {
   position: absolute;
   top: 25px;
   left: 26px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,82 @@
...
+  position: absolute;
+  top: 25px;
+  left: 26px;
+  width: 19px;
+  height: 19px;
+}
+
+.searchbar {
+  width: 100%;
+  font-family: Avenir, sans-serif;
```
Viktor-Kost, line 42: move to `html` tag and use inherit here
```suggestion
  font-family: inherit;
```

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+  border-radius: 3px;
+  font-weight: 900;
+  filter: drop-shadow(0 4px 4px rgba(0, 0, 0, 0.25));
+}
+
+.search.type-1 .searchbar {
+  padding: 23px 10px 23px 62px;
+}
+
+.search.type-2 .searchbar {
```
Viktor-Kost, line 67: if you add modifiers you will use like this:
```suggestion
.form__searchbar--big {
  padding: 23px 10px 23px 62px;
}

.form__searchbar--small {
```

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-weight: 900;
+}
+
+* {
+  box-sizing: border-box;
+  outline: none;
```
Viktor-Kost, line 19: it's allowed to use only `box-sizing: border-box;` at `*` selector

Viktor-Kost (changes requested):

PR #3303, opened 2023-05-25
added task solution for airbnb bar

LabPetro:
[DEMO LINK](https://LabPetro.github.io/layout_search-bar-airbnb/)

[TEST REPORT LINK](https://LabPetro.github.io/layout_search-bar-airbnb/report/html_report/)


src/index.html
```diff
@@ -11,6 +11,48 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="http://localhost:8080/"
```
Anastasiia-Svintsova, line 15: Don't use localhost for the action attribute. You can use this link for now https://mate-academy-form-lesson.herokuapp.com/create-application

src/index.html
```diff
@@ -11,6 +11,48 @@
...
+      class="formBar"
+      data-qa="big"
+    >
+      <label>
+        <img
+          class="icon icon--big"
+          src="images/Search.svg"
+          alt="Logo image"
+        >
+        <input
```
Anastasiia-Svintsova, line 26: Don't forget to add empty lines between multiline sibling blocks of HTML
```suggestion
        >
        
        <input
```

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  font-weight: 300;
+  padding-left: 33px;
+  height: 42px;
+  font-size: 14px;
+}
+
+.small--searchBar::placeholder {
+  font-family: inherit;
+  color: #3d4e61;
+}
```
Anastasiia-Svintsova, line 74: Follow the checklist, please. Don't repeat the same styles 
image

Anastasiia-Svintsova (changes requested):
- Remove this outline. It doesn't match the design 
image

- Fix styles for the entered text on focus. It should be bolder 
image
image

src/style.css
LabPetro: Hi, just one clarification on this one. So you mean i need to create only one main class and then a separate class only for the small one? For example one class="searchbar" and then only another one class="small--searchbar" and put the styles accordingly? Does the same go for the icon classes i created? * Can you please give me a hint if i only need to use outline:none and font weight: 800; only on the focus input? , Sorry i didn't have the option to comment on the first suggested change*.

LabPetro: @Anastasiia-Svintsova

anstsot (approved):
Great job, approved!

PR #3306, opened 2023-06-12
Test Code 

steveforde:


src/index.html
```diff
@@ -1,16 +1,42 @@
...
+            data-qa="keypress"
+            >
+            <button type="submit" class="search-form__button">
+            </button>
+          </label>
+        </div>
+      </form>
+    </div>
+
+    <div class="search-form search-form_small" data-qa="small">
```
Anastasiia-Svintsova, line 26: Same comments go for this form.
It's generally recommended to keep the HTML structure as simple and concise as possible, using wrappers and containers only when necessary to achieve the desired functionality and styling. Using redundant wrappers is not recommended because it can lead to unnecessary complexity and bloated markup.
So, it'll be enough to use this structure for each form
```
<form>
  <label>
    <input>
  </label>
</form>

File: src/style.css [hunk 2]
```diff
@@ -0,0 +1,148 @@
...
+  height: 70px;
+}
+
+.search-form_small .search-form__input {
+  height: 42px;
+}
+
+.search-form__container {
+  position: relative;
+  width: 1050px;
```
Anastasiia-Svintsova, line 62: Don't hardcode the width, search bar must have width 100%

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+}
+
+.search-form__label {
+  display: flex;
```
Anastasiia-Svintsova, line 66: You don't really need to use flex in this task. Will be better to use absolute position for the search image

Anastasiia-Svintsova (changes requested):
Keep up the good job!

src/index.html
```diff
@@ -1,16 +1,32 @@
...
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="search-form search-form_big" data-qa="big">
+      <label class="search-form__label search-form__label_big">
+        <input type="text" 
+          class="search-form__input search-form__input_big" 
+          placeholder="Try &quot;Los Angeles&quot;" 
+          data-qa="keypress"
+        >
```
qoa11a, line 16:
```suggestion
        <input 
          type="text" 
          class="search-form__input search-form__input_big" 
          placeholder="Try &quot;Los Angeles&quot;" 
          data-qa="keypress"
        >
```
Follow the correct code style.
You can find the rules here: https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md

image

src/index.html
```diff
@@ -1,16 +1,32 @@
...
+      </label>
+    </form>
+
+    <form class="search-form search-form_small" data-qa="small">
+      <label class="search-form__label search-form__label_small">
+        <input type="text" 
+          class="search-form__input search-form__input_small" 
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
```
qoa11a, line 25:
```suggestion
        <input 
          type="text" 
          class="search-form__input search-form__input_small" 
          placeholder="Try &quot;Los Angeles&quot;"
        >
```

src/index.html
```diff
@@ -1,16 +1,32 @@
...
+      </label>
+    </form>
   </body>
 </html>
+
+
+
```
qoa11a, line 32: Remove empty rows.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+  font-style: normal;
+  font-weight: 800;
+}
+
+.search-form {
+  position: relative;
+  left: 5px;
+  margin-top: 15px;
+  width: 100%;
+  top: 3px;
```
qoa11a, line 30: You don't need to use `top` and `left` here. You should achieve correct placing by using `margin`.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+.search-form__input {
+  height: 42px;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  color: #3d4e61;
+  font-weight: 300;
+  font-family: "Avenir", sans-serif;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  outline: none;
+  padding-left: 15px;
```
qoa11a, line 42: Why do you set it to `15px`? It's not the correct value for any of the inputs. Plus, you just rewrite it in other selectors.
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+  padding-right: 43px;
+  font-size: 16px;
+  line-height: 22px;
```
qoa11a, line 45: It's better to use relational values for the `line-height` so when you change the `font-size` the `line-height` would change as well.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 800;
+}
+
+.search-form {
+  position: relative;
```
qoa11a, line 26: You don't need `position: relative` for `search-form` as you position the icon with the help of the `label`.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+.search-form__label {
+  position: relative;
+}
+
+.search-form__label:after {
+  content: "";
+  position: absolute;
+  top: 50%;
+  transform: translateY(-50%);
+  left: 16px;
```
qoa11a, line 59: You could achieve the right result with `top: 25px` and `left: 26px` for the big input and `top: 15px` and `left: 13px` for the small one. Please, learn the design carefully.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+  left: 26px;
+  width: 19px;
+  height: 19px;
+  border-radius: 0;
+  box-sizing: border-box;
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  background-position: top left top;
+}
```
qoa11a, line 96: Move similar properties under the same selector to not repeat the same code.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+
+.search-form_big {
+  margin-left: -5px;
+  margin-top: 15px;
+  position: relative;
+  border: 1px solid #e1e7ed;
+  background: linear-gradient(#e1e7ed, #e1e7ed), linear-gradient(#fff, #fff);
+  background-origin: border-box;
+  background-clip: content-box, border-box;
+}
```
qoa11a, line 129: Please, don't use random properties and random values. I don't know where you've taken them from. Just apply styles from the design.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+
+.search-form__input_big::placeholder {
+  height: 22px;
+  width: 129px;
+}
```
qoa11a, line 134: It's incorrect to set the width and height of the placeholder. It should be set automatically depending on the input itself.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+
+.search-form__input_small::placeholder {
+  height: 20px;
+  width: 167.29px;
```
qoa11a, line 138: Same here.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+  font-size: 14px;
+  font-weight: 300;
+  line-height: 19px;
+  padding-left: 0;
+  margin-left: -5px;
```
qoa11a, line 143: Position it width `padding-left` for the input itself.

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+}
+
+.search-form_small .search-form__label:after {
+  height: 11px;
+  width: 11px;
+  left: 13px;
+  top: 1px;
+}
```
qoa11a, line 151: Why do you rewrite styles for the same selector a few times?

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+
+.search-form__label_small:after {
+  left: 13px;
+  transform: translateY(6px);
+}
```
qoa11a, line 156: What is it?

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+
+.search-form__input_small {
+  padding-left: 30px;
+  padding-top: 3px;
```
qoa11a, line 160: Not the right value. It looks okay because you've set the height for the placeholder. But it's incorrect to do so.

src/index.html
polosanya: Not fixed

src/style.css
```diff
@@ -0,0 +1,165 @@
...
+  line-height: 22px;
+  width: 100%;
+  box-sizing: border-box;
+}
+
+.search-form__label {
+  position: relative;
+}
+
+.search-form__label:after {
```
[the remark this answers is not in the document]
polosanya: Not fixed

polosanya (approved):
Nice! I'm approving it, but please make sure that you fixed all previous comments 🙂

steveforde:
Thanks
Sorry I missed that
I update now

PR #3312, opened 2023-06-21
add search bar airbnb

daria-sazon:
- [DEMO LINK](https://daria-sazon.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://daria-sazon.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Search bar for Airbnb
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_search-bar-airbnb/)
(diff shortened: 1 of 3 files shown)


Viktor-Kost (changes requested):
You have repeatable styles. 

Following the BEM-methodology - use the `block` selector for the FIRST form (`class="form"`), where you apply common form styles

```
.form {
  margin: 20px 0;
  border: solid 1px #e1e7ed;
  border-radius: 4px;
  height: 70px;
  position: relative;
  margin-top: 20px;
}
```

And then you apply `modifier` to the SECOND form (`class="form form--small"`) that have ONLY differences with base class form
```
.form--small {
  height: 42px;
  border-radius: 4px;
}
```
So in this way you follow DRY (Don't repeat yourself) principle

IvanFesenko (approved):

PR #3315, opened 2023-06-22
add task solution

Zaransoftline:
https://Zaransoftline.github.io/layout_search-bar-airbnb/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,117 @@
+* {
+  padding: 0;
+  box-sizing: border-box;
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
+  
+.form {
+  display: flex;
+  justify-content: center;
+  margin: 20px auto 0;
+}
+  
+.form--big {
+  width: 100%;
+  height: 70px;
+}
+  
+.form--small {
+  width: 100%;
+  height: 42px;
+}
+  
+.form-label {
+  position: relative;
+  display: block;
+  width: 100%;
+}
+  
+.form-label:after {
+  position: absolute;
+  content: "";
+  display: block;
+}
+  
+.form_label-big:after {
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 19px;
+  height: 19px;
+  top: 25px;
+  left: 26px;
+}
+  
+.form_label-small:after {
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 11px;
+  height: 11px;
+  top: 15px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,117 @@
...
+  font-style: normal;
+  src: url("fonts/Avenir-Book.ttf") format("truetype");
+}
+    
+@font-face {
+  font-family: "Avenir";
+  font-weight: 600;
+  font-style: normal;
+  src: url("fonts/Avenir-Heavy.ttf") format("truetype");
+}
```
anastasiiavorobiova, line 117: Font faces should be on the top of your file

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+  
+.form {
+  display: flex;
+  justify-content: center;
+  margin: 20px auto 0;
+}
+  
+.form--big {
+  width: 100%;
+  height: 70px;
```
anastasiiavorobiova, line 16: The height property should be defined by content height, we shouldn't set it in this case

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+}
+  
+.form--small {
+  width: 100%;
+  height: 42px;
```
anastasiiavorobiova, line 21: The height property should be defined by content height, we shouldn't set it in this case

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+  background-size: contain;
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
+  
+.input {
+  width: 100%;
+  height: 100%;
```
anastasiiavorobiova, line 58: The height property should be defined by content height, we shouldn't set it in this case

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+  content: "";
+  display: block;
+}
+  
+.form_label-big:after {
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 19px;
+  height: 19px;
```
anastasiiavorobiova, line 41: The height property should be defined by content height, we shouldn't set it in this case

anastasiiavorobiova (changes requested):
Great start! Added comments to fix

src/style.css
```diff
@@ -0,0 +1,115 @@
+@font-face {
+  font-family: "Avenir";
+  font-weight: 300;
+  font-style: normal;
```
volodymyr-soltys97, line 4: This is default value, you can remove it 
```suggestion
```

volodymyr-soltys97 (approved):
Good work 👍

PR #3331, opened 2023-06-29
Search bar AirBNB

aRabiej:
- [DEMO LINK](https://aRabiej.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://aRabiej.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,7 +10,29 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
     <h1>Search bar airbnb</h1>
+    <form class ="search-form search-form_big" data-qa="big">
+      <label class ="search-form__label search-form__label_big">
+        <input
+          data-qa ="keypress"
+          class= "search-form__input  search-form__input_big"
+          type ="text"
+          placeholder ="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
+
+    <form class ="search-form search-form_small" data-qa="small">
+      <label class ="search-form__label search-form__label_small">
+        <input
+          data-qa ="hover"
+          class ="search-form__input search-form__input_small"
+          type ="text"
+          placeholder ="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -12,7 +12,6 @@
   </head>
 
   <body>
-    <h1>Search bar airbnb</h1>
     <form class ="search-form search-form_big" data-qa="big">
       <label class ="search-form__label search-form__label_big">
         <input
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,120 @@
+* {
+  padding: 0;
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 300;
+  font-style: normal;
+  src: url(./fonts/Avenir-Book.ttf) format("truetype");
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 600;
+  font-style: normal;
+  src: url(./fonts/Avenir-Heavy.ttf) format("truetype");
+}
+
+body {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
+
+.search-form {
+  display: flex;
+  justify-content: center;
+  margin: 20px auto 0;
+}
+
+.search-form_big {
+  width: 100%;
+  height: 70px;
+}
+
+.search-form_small {
+  width: 100%;
+  height: 42px;
+}
+
+.search-form__label {
+  position: relative;
+  display: block;
+  width: 100%;
+}
+
+.search-form__label:after {
+  position: absolute;
+  content: "";
+  display: block;
+}
+
+.search-form__label_big:after {
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 19px;
+  height: 19px;
+  top: 25px;
+  left: 26px;
+}
+
+.search-form__label_small:after {
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
+
+.search-form__input {
+  width: 100%;
+  height: 100%;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  color: #3d4e61;
+  font-weight: 300;
+  font-family: inherit;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  outline: none;
+}
+
+.search-form__input:hover {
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -10,7 +10,28 @@
...
+          type ="text"
+          placeholder ="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
+
+    <form class ="search-form search-form_small" data-qa="small">
+      <label class ="search-form__label search-form__label_small">
+        <input
+          data-qa ="hover"
```
darokrk, line 29: The data-qa attribute for the small input should be `data-qa="small`

image

src/index.html
```diff
@@ -10,7 +10,28 @@
...
+          data-qa ="keypress"
+          class= "search-form__input  search-form__input_big"
+          type ="text"
+          placeholder ="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
+
+    <form class ="search-form search-form_small" data-qa="small">
+      <label class ="search-form__label search-form__label_small">
```
darokrk, line 27: Same here add `for` attribute

src/index.html
```diff
@@ -10,7 +10,28 @@
...
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class ="search-form search-form_big" data-qa="big">
+      <label class ="search-form__label search-form__label_big">
+        <input
+          data-qa ="keypress"
+          class= "search-form__input  search-form__input_big"
+          type ="text"
+          placeholder ="Try &quot;Los Angeles&quot;"
+        >
```
darokrk, line 22: Don't forget to close the input tag.

src/index.html
```diff
@@ -10,7 +10,28 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class ="search-form search-form_big" data-qa="big">
+      <label class ="search-form__label search-form__label_big">
```
darokrk, line 16: It's recommended to add `for` attribute in `label` tag for better accessibility. The `for` attribute specifies which form element a label is bound to. It should be the same value as you have on input `id`

src/index.html
```diff
@@ -10,7 +10,28 @@
...
+        <input
```
darokrk, line 17: It's recommended to add `id` and `name` attributes to input to avoid weird behavior and make sure the data is always sent correctly

src/index.html
```diff
@@ -10,7 +10,28 @@
...
+      </label>
+    </form>
+
+    <form class ="search-form search-form_small" data-qa="small">
+      <label class ="search-form__label search-form__label_small">
+        <input
+          data-qa ="hover"
+          class ="search-form__input search-form__input_small"
+          type ="text"
+          placeholder ="Try &quot;Los Angeles&quot;"
```
darokrk, line 32: Please add as above `id` and `name` attributes

src/index.html
```diff
@@ -10,7 +10,28 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class ="search-form search-form_big" data-qa="big">
```
darokrk, line 15: This form should have just `class = search-form`, same for label `class ="search-form__label` and input `class= "search-form__input`

For small one you just add a modifier class `<form class ="search-form search-form--small"`
For small form label `<label class ="search-form__label search-form__label--small">`
For small form input `class ="search-form__input search-form__input--small"`

src/style.css
```diff
@@ -0,0 +1,120 @@
...
+  font-style: normal;
+  src: url(./fonts/Avenir-Heavy.ttf) format("truetype");
+}
+
+body {
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+}
+
+.search-form {
```
darokrk, line 25: add `width: 100%` here so you would not need it in modifier class `.search-form--small`

darokrk (changes requested):
UI looks good 😃 few improvements are needed and we got it 👍🏽

src/index.html
```diff
@@ -10,7 +10,32 @@
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
```
Radoslaw-Czerniawski, line 13: Remove this empty line, it's unnecessary

src/index.html
```diff
@@ -10,7 +10,32 @@
...
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class ="search-form" data-qa="big">
```
Radoslaw-Czerniawski, line 15: You have a lot of misformatted attributes with spaces between them -> `class = ` this is not correct, it should be `class=` correct all of such attributes

src/index.html
Radoslaw-Czerniawski: input here is fine, habit from React on Darek's part most likely (in jsx you add the closing tag for inputs)

Radoslaw-Czerniawski (changes requested):
Generally the pr looks very good, adjust according to the comments and also add the shadows underneath text that are present in the figma design:
#### figma
image

#### pr
image

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+
+.search-form__label--small:after {
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
```
DorotaLeniecDev, line 64: You dont need to repeate the common styles added for **.search-form__label**
```css
.search-form__label--small:after {
  width: 11px;
  height: 11px;
  top: 15px;
  left: 13px;
}
```

DorotaLeniecDev (changes requested):
Demo and test pages are not working, could you check that?

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+.search-form__label:after {
+  position: absolute;
+  content: "";
+  display: block;
+  background-image: url(./images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 19px;
+  height: 19px;
+  top: 25px;
```
Radoslaw-Czerniawski, line 52: I assume these values are not halfs of the height and padding to counter :after 19x19 size. This will be hard to work with as you have to manually calculate images position. A better alternative would be to position base of the :after with half of these values and then translate it by 50% of the images size -> this way you will more accurately position it in the center.
```css
top: 35px;
left: 31px;
transform: translate(-50%, -50%) 
```
or you could even use some variables as these values are dependent on each other
```css
.search-form {
  height: var(--big-form-height);
  ...
}

.search-form__label:after {
  top: calc(var(--big-form-height) / 2);
  left: calc(var(--big-input-padding-l) / 2);  /* -> same for the padding */
  transform: translate(-50%, -50%);
  ...
}
```

Radoslaw-Czerniawski (approved):
Looks exactly like in the Figma design, good job :D
Few notes:
- don't commit everything, only add relevant changes to the pr (here changes in the dist folder are not needed and are polluting review)
- left a comment with some cool solution for the `:after` position, it may come in handy

PR #3369, opened 2023-07-07

thaiduk:
- [DEMO LINK](https://thaiduk.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://thaiduk.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,56 +1,84 @@
-.search__bar {
-  position: relative;
-  display: flex;
-  align-items: center;
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+  font-style: normal;
 }
 
-.search__bar--img {
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 600;
+  font-style: normal;
+}
+
+* {
+  margin: 0;
+}
+
+.form__container {
+  margin: 20px auto 0;
+}
+
+.search-form__label {
+  box-sizing: border-box;
+  display: block;
+  position: relative;
+}
+
+.search-form__label:after {
   width: 19px;
   height: 19px;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  content: "";
+  display: block;
   position: absolute;
+  top: 25px;
   left: 26px;
 }
 
-.search__bar--field {
-  line-height: 70px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,56 @@
...
+  left: 13px;
+}
+
+.search__barSmall--field {
+  width: 427px;
+  height: 42px;
+  line-height: 42;
+  font-size: 14px;
+  padding-left: 33px;
+}
```
alexander-ignatow, line 56: You already have `search__bar--small` style on the wrapper div, so you can use it to alter the styles of `search__bar--img` and `search__bar--input`. In fact, it will remove the need for additional classes (`search__barSmall--img` and  `search__barSmall--field`)

alexander-ignatow (commented):
The following requirements are not met:

* search bar must have width 100%
* distance between two search bars must be 20px
* default font-weight must be 300 (see more comments in the code)
* add attribute data-qa="big" for big search form, and data-qa="small" for small
* add attribute data-qa="keypress" to input in big search form

src/style.css
```diff
@@ -0,0 +1,56 @@
...
+  left: 26px;
+}
+
+.search__bar--field {
+  line-height: 70px;
+  color: #3d4e61;
+  font-family: Avenir, sans-serif;
+  font-size: 16px;
+  font-style: normal;
+  font-weight: 350;
```
alexander-ignatow, line 20: Font weight value `350` is not valid with modern browsers
See more https://www.w3schools.com/cssref/pr_font_weight.php

BTW - now you have good experience to see how browsers handle weird font-weights)

alexander-ignatow (changes requested):
Overal - not bad for the first try.
Visually everything is very close to what we need

However - we need some more efforts to polish everything

First:
Check following checklist items that are not implemented:
- Add ALL data-qa attributes required in the task
- Text Try "Los Angeles" is a placeholder.
- Use &quot; instead of " in the placeholder.
- You have to put input inside the form for correct data processing. Form should have action and method attributes.
- Each search bar to be inside its own form element.
- Task has styling for :focus state. Forms, labels, divs are not focusable by default. Consider it when selecting your layout. (pay attention here - focused styling is not applied at al in your case)

thaiduk:
I've rewrited some items in code and as result make simplest styles. Let's check. Will be thankful for your comments.

roman-mirzoian (changes requested):
Well done, but now not all of your tests pass.
image

PR #3385, opened 2023-07-12
add search bar

FrontKid:
- [DEMO LINK](https://FrontKid.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://FrontKid.github.io/layout_search-bar-airbnb/report/html_report/)


src/index.html
```diff
@@ -2,15 +2,45 @@
...
+      action="#"
+      method="post"
+      data-qa="small"
+      >
+        <label class="search-form__lable search-form__lable--small">
+          <input
+            class="search-form__field search-form__field--sm"
+            type="text"
+            placeholder="Try &quot;Los Angeles&quot;"
+            >
```
roman-mirzoian, line 41: The small search bar input field is missing the `data-qa` attribute.

src/style.css
```diff
@@ -0,0 +1,98 @@
...
+  font-style: normal;
+  font-weight: 300;
+  line-height: normal;
+}
+
+body {
+  margin: 0 8px;
+}
+
+.search-form__lable {
```
roman-mirzoian, line 28: Be careful when naming classes.
There is a typo in the class name. It should be `search-form__label` instead of `search-form__lable`.

src/style.css
```diff
@@ -0,0 +1,98 @@
...
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  transition: box-shadow 0.3s ease 0s;
+  box-sizing: border-box;
+}
+
+.search-form__field:hover {
+  box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.25);
+}
+
+.search-form__field::placeholder:focus,
```
roman-mirzoian, line 63: The `search-form__field::placeholder:focus` selector is invalid. CSS doesn't provide any way to style the placeholder on focus.

src/style.css
```diff
@@ -0,0 +1,98 @@
...
+
+.search-form__field {
+  display: block;
+  font-family: inherit;
+  width: 100%;
+  outline: none;
+  border-radius: 4px;
+  color: inherit;
+  font-size: inherit;
+  font-weight: inherit;
```
roman-mirzoian, line 51: Do we really need to manually specify styles with the `inherit` value here? CSS styles are automatically followed.

src/style.css
```diff
@@ -0,0 +1,98 @@
+@font-face {
+  font-family: "Avenir";
```
roman-mirzoian, line 2: It is better to put the values of font-family, colors and sizes into separate variables and reuse them.

roman-mirzoian (changes requested):
Good work, but it needs to be improved a little 👍

src/style.css
FrontKid: if i remove these rules the styles aren't followed except font-weight so i removed it

src/index.html
```diff
@@ -2,15 +2,44 @@
...
   <body>
-    <h1>Search bar airbnb</h1>
+    <section class="search">
+      <form
+      class="search-form search__bar--lg"
+      action="#"
+      method="post"
+      data-qa="big"
+      >
+
```
roman-mirzoian, line 19: Don't forget about good formatting and don't leave extra empty lines.

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+  margin-top: 20px;
+  margin-bottom: 20px;
+}
+
+.search-form__label--large::before {
+  width: 19px;
+  height: 19px;
+  left: 26px;
+  top: 26px;
+
```
roman-mirzoian, line 82: Don't forget about good formatting and don't leave extra empty lines.

src/index.html
```diff
@@ -2,15 +2,44 @@
...
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <section class="search">
+      <form
+      class="search-form search__bar--lg"
+      action="#"
+      method="post"
+      data-qa="big"
+      >
```
roman-mirzoian, line 18: Fix it in the whole file.
image

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/index.html
FrontKid: done

src/index.html
roman-mirzoian: Why did you react with thumb, but not delete the extra blank line?)

src/style.css
```diff
@@ -0,0 +1,98 @@
...
+.search-form__label--small::before {
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
+
+.search-form__field--sm {
+  padding: 10.2px 33px;
```
roman-mirzoian, line 96: If you specify fractional sizes, then something is done incorrectly.

src/style.css
```diff
@@ -0,0 +1,98 @@
+:root {
+  --avenir-family: "Avenir";
```
roman-mirzoian, line 2: It is better to name variables not by content, but by value, because the content can change during the development process, so it would be better to name it `--main-font-family`

src/style.css
```diff
@@ -0,0 +1,98 @@
...
+}
+
+@font-face {
+  font-family: "Avenir";
```
roman-mirzoian, line 6: You created a variable for this parameter, so you should use it wherever possible.

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/style.css
FrontKid: if I use it in font face it doesn't work at all

roman-mirzoian (approved):
Well done 👍

PR #3394, opened 2023-07-14
Develop

VitaliiNez:
https://vitaliinez.github.io/layout_search-bar-airbnb/
https://vitaliinez.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,12 @@
+:root {
+  --main-color: #3d4e61;
+  --border-color: #e1e7ed;
+  --shadow-black: #00000040;
+  --shadow-light: #3d4e611a;
+  --white-color: #fff;
+  --light-color: #f6f6f7;
+}
+
 * {
   box-sizing: border-box;
   margin: 0;
@@ -24,13 +33,13 @@ body {
 }
 
 input {
-  background: #fff;
-  border: 1px solid #e1e7ed;
-  box-shadow: 0 1px 8px 0 rgba(61,78,97,0.1);
+  background: var(--white-color);
+  border: 1px solid var(--border-color);
+  box-shadow: 0 1px 8px 0 var(--shadow-light);
   border-radius: 4px;
   outline: none;
   line-height: 21.86px;
-  color: #3d4e61;
+  color: var(--main-color);
   width: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -11,6 +11,41 @@
...
+            <input
+              class="form__big-input"
+              type="search"
+              name="City-small"
+              id="big-form"
+              placeholder="Try &quot;Los Angeles&quot;"
+            >
+          </label>
+        </form>
+        <form
```
Anastasiia-Svintsova, line 32: Don't forget to add empty lines between multiline sibling blocks of HTML
```suggestion
        </form>
        
        <form
```

src/style.css
```diff
@@ -0,0 +1,132 @@
+*,
+*::after,
+*::before {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
```
Anastasiia-Svintsova, line 7: It is not considered the best practice to apply the reset styles to all elements using the `*` selector. It is generally recommended to apply reset styles selectively, targeting only the specific elements that need them. You can leave only box-sizing here to set this property uniformly across all elements
```suggestion
* {
  box-sizing: border-box;
}
```

src/style.css
```diff
@@ -0,0 +1,132 @@
...
+
+body {
+  font-family: "Avenir", sans-serif;
+  margin: 0;
+}
+
+input {
+  font-family: "Avenir", sans-serif;
+}
```
Anastasiia-Svintsova, line 30: You/ve already set the font family inside the body tag. So, remove it

src/style.css
```diff
@@ -0,0 +1,132 @@
...
+  font-weight: 300;
+  color: #3d4e61;
+}
+
+.form__big-label {
+  position: relative;
+  display: block;
+}
+
+.form__big-label::before {
```
Anastasiia-Svintsova, line 74: Follow the checklist, please. Don't repeat the same styles 
image

Anastasiia-Svintsova (changes requested):
- Fix the font style according to the mock when the input is in focus 
image

- Remove this button when the input is in focus 
image

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+
+body {
+  font-family: "Avenir", sans-serif;
+  margin: 0;
+}
+
+input {
+  background: #fff;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px 0 rgba(61,78,97,0.1);
```
OleziO, line 29: Use HEX instead of RGBA.

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+  border-radius: 4px;
+  outline: none;
+  line-height: 21.86px;
+  color: #3d4e61;
```
OleziO, line 33: Move all repeatable colors to the CSS variables.

OleziO (changes requested):

anastasiiavorobiova (approved):
Really nice!

PR #3406, opened 2023-07-18
add task solution

valet2102:


src/style.css
```diff
@@ -0,0 +1,99 @@
...
+  src: url(./fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url(./fonts/Avenir-Heavy.ttf);
+  font-weight: 700;
+}
+
```
roman-mirzoian, line 13: Do not leave extra empty lines.

src/style.css
```diff
@@ -0,0 +1,99 @@
...
+/* #endregion fonts */
+
+/* #region base-styles */
+input {
```
roman-mirzoian, line 17: It is better to tie stylizations more to classes and work directly with tags as little as possible.

roman-mirzoian (approved):
Well done 👍

PR #3408, opened 2023-07-18
add task solution

kaavka:


src/index.html
```diff
@@ -10,7 +10,51 @@
...
+    <form
+      class="search"
+      data-qa="big"
+      action="/action/api"
+      method="post"
+    >
+      <label for="searchInputBig">
+        <img
+          src="images/Search.svg"
+          alt="search"
```
roman-mirzoian, line 24: The `alt` attribute should be more meaningful.

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+.visually-hidden {
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  margin: -1px !important;
+  border: 0 !important;
+  padding: 0 !important;
+  overflow: hidden !important;
+  clip: rect(0,0,0,0) !important;
+  white-space: nowrap !important;
```
roman-mirzoian, line 28: Using `!important` - says that you are doing something wrong, it can be used only in extreme cases.

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

Also, pay attention to the appearance of your application and the test results:
image

src/style.css
kaavka: I've just thought that h1 is necessary in html document. And so i just hide it with bootstrap's visualy hiden class. I know that its unnecessary right here but is this wrong? @roman-mirzoian

roman-mirzoian: @kaavka https://w3schoolsua.github.io/css/css_important.html#gsc.tab=0

roman-mirzoian (approved):
It's good that you did extra work, but don't forget to correct comments from mentors and try to pass your tests. For them to pass, it was enough to change the values ​​in the styles a little.

PR #3425, opened 2023-07-22
'v1'

DVasylenko18:
- [DEMO LINK](https://DVasylenko18.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://DVasylenko18.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,40 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="search search_big" data-qa="big">
+      <div class="search_big__line">
+        <img
+          type="submit"
+          class="icon search_big__icon"
+          src="./images/Search.svg"
+          alt="search"
+        >
+        <input
+          data-qa="keypress"
+          type="text"
+          class="search__input search_big__input"
+          name="searchbar"
+          placeholder="Try &quot;Los Angeles&quot"
+        >
+      </div>
+    </form>
+
+    <form class="search search_small" data-qa="small">
+      <div class="search_small__line">
+        <img
+          type="submit"
+          class="icon search_small__icon"
+          src="./images/Search.svg"
+          alt="search"
+        >
+        <input
+          data-qa="keypress"
+          type="text"
+          class="search__input search_small__input"
+          name="searchbar"
+          placeholder="Try &quot;Los Angeles&quot"
+        >
+      </div>
+    </form>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,40 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="search search_big" data-qa="big">
+      <div class="search_big__line">
```
xiimiik, line 15: let's simplify the code and remove this wrapper along with its styles

src/style.css
```diff
@@ -0,0 +1,98 @@
...
+  font-style: inherit;
+  font-weight: inherit;
+  width: 100%;
+  height: 90%;
+  border: 0;
+  outline: 0;
+  background: none;
+}
+
+.search:focus-within {
```
xiimiik, line 62: after removing the extra wrappers, you can use a simple `:focus'

src/style.css
```diff
@@ -0,0 +1,98 @@
...
+}
+
+.search_big__icon {
+  display: inline-block;
+  width: 19px;
+  height: 19px;
+  padding-right: 18px;
+}
+
+.search__input {
```
xiimiik, line 49: this should be our general class, in addition, there should be only a class for it that will slightly change the styles for the small search bar

src/style.css
```diff
@@ -0,0 +1,98 @@
...
+  display: flex;
+  justify-content: left;
+  align-items: center;
+  height: 68px;
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  background: #fff;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  margin: 20px auto 0;
+}
```
xiimiik, line 28: As I've described below, we don't need to separate styles and give unnecessary wrappers, so this class should only have indents or `position: relative' if the image is conditionally positioned with an absolute. You can also set the image as a background

xiimiik (changes requested):
GJ! Keep going!

src/index.html
DVasylenko18: done

src/style.css
DVasylenko18: even after remoming still doesn't work with simple focus

src/style.css
DVasylenko18: done

src/style.css
DVasylenko18: moved to search__input class

Esceype (approved):
Great job!!!

PR #3436, opened 2023-07-23
add task solution

pzwolenik:
- [DEMO LINK](https://pzwolenik.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://pzwolenik.github.io/layout_search-bar-airbnb/report/html_report/)


DorotaLeniecDev (changes requested):
DPY - dont repeat yourself, refactor css to have common class for styling and a modifier to add differences. Also make sure input hover and focus state have proper styles - for now they are dirrefent then in the Figma design. Also make sure placeholder styles are not changing, only input text styles should be different

mykhalenych (approved):

PR #3454, opened 2023-07-29
search

Dmytro-Kulyk-cpu:


src/style.css
```diff
@@ -0,0 +1,86 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  font-family: "Avenir", sans-serif;
+}
+/*
+.container {
+  margin-top: 20px;
```
volodymyr-soltys97, line 12: Remove unused styles

volodymyr-soltys97 (changes requested):
Add the indents on left and right and try to fix problem with test 
image

etojeDenys (approved):
good job

PR #3457, opened 2023-07-29
add task solution

vinogradova8:
- [DEMO LINK](https://vinogradova8.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://vinogradova8.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,90 @@
+@font-face {
+  font-family: "Avenir";
+  src: url("fonts/Avenir-Book.ttf");
+  font-style: normal;
+  font-weight: 400;
+}
+
+@font-face {
+  font-family: "Avenir";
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  position: absolute;
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+.icon--big {
+  left: 26px;
+  width: 19px;
+  height: 19px;
+}
```
xiimiik, line 37: no need to give the modifier `--big`, just add it for the icon, the modifier is needed only for the small one

xiimiik (changes requested):
color for input should be another
image
image

xiimiik (approved):
GJ🔥 🚒

PR #3460, opened 2023-07-30
Develop

cty2802:


src/style.css
```diff
@@ -0,0 +1,60 @@
...
+}
+
+.search-bar__input {
+  width: 100%;
+  font-size: 16px;
+  box-sizing: border-box;
+  padding: 25px 0 25px 62px;
+  box-shadow: var(--input-box-shadow);
+  background: var(--input-bg-color)
+    url(./images/Search.svg) center left 27px / 19px 19px no-repeat;
```
Anna-Tkachenko, line 32: It works with current sizes, but won't work if the size of the input will be changed
Will be great to think how we can calculate this position according to the size of the input

src/style.css
```diff
@@ -0,0 +1,60 @@
...
+  --input-border-color: #e1e7ed;
+  --input-bg-color: #fff;
+  --input-box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  --input-box-shadow-hover: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+  --input-bg-focus: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  --placeholder-color: #3d4e61;
+}
+
+@font-face {
+  font-family: "Avenir";
```
Anna-Tkachenko, line 11: Font styles should be added to html tag

Anna-Tkachenko (approved):

PR #3469, opened 2023-08-01
add task solution

Xsaites:


src/style.css
```diff
@@ -0,0 +1,94 @@
...
+  font-size: normal;
+}
+
+@font-face {
+  font-family: "Avenir", Helvetica, Arial, sans-serif;
+  src: url(fonts/Avenir.ttc);
+  font-display: swap;
+  font-weight: 400;
+  font-size: normal;
+}
```
Esceype, line 23: Remove it and import create with font-weight 900

src/index.html
```diff
@@ -1,16 +1,49 @@
...
+        class="form__line"
+      >
+        <input
+          data-qa="keypress"
+          type="text"
+          class="form__input" placeholder="Try &quot Los Angeles &quot "
+        >
+        <div class="form__search">
+          <img src="images/Search.svg" alt="search icon">
+        </div>
```
Esceype, line 30: Move it to top before input, will be better for understanding of code styles, remove div and add class to img tag

Esceype (changes requested):
Overall good job! Make these changes and let`s try to pass the tests after the fixes. All what I wrote about big form the same for small form

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+
+@font-face {
+  font-family: "Avenir", Helvetica, Arial, sans-serif;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-display: swap;
+  font-weight: 900;
+  font-size: normal;
+}
+
+body {
```
Esceype, line 17: Reset default browser margins

Esceype (approved):
Great job!!! Just don`t forget to reset default browser margins in body selector

PR #3513, opened 2023-08-15
add task solution

wsadocha:
- [DEMO LINK](https://wsadocha.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://wsadocha.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,94 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 700;
+}
+
+:root {
+  --light-shadow: rgba(61, 78, 97, 0.1);
+  --dark-shadow: rgba(61, 78, 97, 0.2);
+  --gradient: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: "Avenir", Verdana, sans-serif;
+}
+
+.form {
+  display: flex;
+  width: 100%;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,94 @@
...
+
+.form__input--small::placeholder {
+  font-size: 14px;
+}
+
+.form__icon {
+  width: 19px;
+  height: 19px;
+  position: absolute;
+  top: 25px;
```
Radoslaw-Czerniawski, line 75: Here these values are reliant on the inputs padding and height, you could export them to a separate variable:
```css
.form {
  display: flex;
  width: 100%;
  height: var(--big-form-height);  /* this as well as the padding left could be used as variables */
  margin: 20px 0;
}

...

.form__icon {
  top: calc(var(--big-form-height) / 2);  /* now instead of calculating the values on our own, we may use the calc */
  left: calc(var(--big-padding-left) / 2); /* function and divide these values by two to place icon in the middle */
  transform: translate(-50%, -50%); /* icon positions its edge in the middle so we have to translate it by its size */
  ...
}
```

Radoslaw-Czerniawski (changes requested):
- There should be a shadow underneath the letters, use styles from [figma](https://www.figma.com/file/kf3AWulK9elrNk34wtpjPw/Airbnb-Search-bar?type=design&node-id=5102-478&mode=design&t=adOdxTLQCJ1LwWbw-0)
image

- optionally you could create these variables mentioned in the comment

src/style.css
```diff
@@ -0,0 +1,101 @@
...
+
+.form__input--small::placeholder {
+  font-size: 14px;
+}
+
+.form__icon {
+  width: 19px;
+  height: 19px;
+  position: absolute;
+  /* top: 25px; */
```
mykhalenych, line 79: delete comments

mykhalenych (approved):

PR #3519, opened 2023-08-17
add task solution

Daniilart01:
- [DEMO LINK](https://Daniilart01.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Daniilart01.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,9 +8,50 @@
     >
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Search bar airbnb</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form
+      action="http://localhost:8000"
+      method="get"
+      class="search--box big"
+      data-qa="big"
+    >
+      <img
+        src="images/Search.svg"
+        alt="Search"
+        class="big--search--logo"
+      >
+      <label for="searchBoxBig"></label>
+      <input
+        type="text"
+        class="search--bar big"
+        name="searchBoxBig"
+        id="searchBoxBig"
+        placeholder="Try &quot;Los Angeles&quot;"
+        data-qa="keypress"
+      >
+    </form>
+
+    <form
+      action="http://localhost:8000"
+      method="get"
+      class="search--box small"
+      data-qa="small"
+      >
+      <img
+        src="images/Search.svg"
+        alt="Search"
+        class="small--search--logo"
+      >
+      <label for="searchBoxSmall"></label>
+      <input
+        type="text"
+        class="search--bar small"
+        name="searchBoxSmall"
+        id="searchBoxSmall"
+        placeholder="Try &quot;Los Angeles&quot;"
+      >
+    </form>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,113 @@
+body {
+  --logo-big-size: 19px;
+  --logo-small-size: 11px;
+  --big--search-size: 70px;
+  --small--search-size: 42px;
+  --search-field-height: 19px;
+  --text-shadow: 0 4px 4px rgba(0,0,0,0.25);
+}
+
+.search--box {
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  background: #fff;
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  height: var(--big--search-size);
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,113 @@
...
+  text-shadow: var(--text-shadow);
+}
+
+.big--search--logo {
+  width: var(--logo-big-size);
+  height: var(--logo-big-size);
+
+}
+
+.small--search--logo {
```
denys-danyliuk, line 96: 2'nd point in checklist

src/index.html
```diff
@@ -8,9 +8,50 @@
...
+        data-qa="keypress"
+      >
+    </form>
+
+    <form
+      action="http://localhost:8000"
+      method="get"
+      class="search--box small"
+      data-qa="small"
+      >
```
denys-danyliuk, line 41:
```suggestion
    <form
      action="http://localhost:8000"
      method="get"
      class="search--box small"
      data-qa="small"
    >
```
closing bracket should be on the same level as opening one

src/style.css
```diff
@@ -0,0 +1,113 @@
...
+
+@font-face {
+  font-family: "Avenir";
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 400;
+}
+
+@font-face {
+  font-family: "AvenirHeavy";
```
denys-danyliuk, line 109: 7'th point

denys-danyliuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,113 @@
...
+  display: flex;
+  box-sizing: border-box;
+  align-items: center;
+  margin-top: 20px;
+  padding-left: 26px;
+}
+
+.search--box:hover {
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+}
```
xiimiik, line 25: are we really need 2 different hovers? just do only for box or bar

xiimiik (changes requested):

src/style.css
Daniilart01: I have only one `:hover` usage in my code, and it`s code, that u selected, no more

src/style.css
```diff
@@ -0,0 +1,113 @@
...
+
+/*Set up background to container when forms' child (input) focused*/
+.search--box:focus-within {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%) !important;
+}
+
+/*Add shadow to input text in big container when input focused*/
+.search--bar--big:focus.search--bar--big {
+  text-shadow: var(--text-shadow);
+}
```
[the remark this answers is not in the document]
Daniilart01: Can not agree with this comment. I implement figma`s design in all details, and here is description for selectors:
1. This block required to set font-color for both placeholders
```
.search--bar::placeholder {
  color: #3d4e61;
}
```
2.  This adds shadow only into small-bars placeholder ass it's designed in figma (without this rule there will be no shadow in small placeholder and this block is only way to implement this feature)
```
.search--bar--small::placeholder {
  text-shadow: var(--text-shadow);
}
```

3. Default focus properties
```
.search--bar:focus {
  outline: none;
  font-weight: 900;
  /*Because of gradient applying transparent when it's focused*/
  background: transparent;
  border-radius: 3px;
}
```

4. In block 3, when bar focused, we sets up `font-weight: 900;`. As noted in checklist's point 11: "Placeholder shoudn't change style on focus". So this block resets bold font weight for placeholder even if bar focused.
```
.search--bar:focus::placeholder {
  font-weight: normal;
}
```

5. This block adds shadow to input text in big search when it is focused. Separate block because shadow needed only in big search. (I remove `.search--bar--big` from selector cause it is really no need in it)
```
.search--bar--big:focus {
  text-shadow: var(--text-shadow);
}
```

6. Likewise point 4 in this list, this block is required to override shadow that sets up in point 5 of this list (for big search every time it's focused), required because of point 11 in checklist
```
.search--bar--big:focus::placeholder {
  text-shadow: none;
}
```

7. This block is required for shadow removing in small search when focused (big should have shadow even being focused so we need separate selector)
```
.search--bar--small:focus {
  box-shadow: none;
}
```

8. This block required because we shold set background to search--box when focused it's child (search--bar). For these purpose we have `:focus-within`
.search--box:focus-within {
  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%) !important;
}

If u still think I should decrease selectors amount - provide some instructions, cause all of declared selectors are required in my implementation

mykhalenych (approved):

PR #3524, opened 2023-08-19
add task solution

Lidia-Pintos-Medina:


src/style.css
```diff
@@ -0,0 +1,117 @@
...
+}
+
+.text-small::placeholder {
+  font-size: 14px;
+  height: 20px;
+  position: relative;
+  line-height: 19.12px;
+  margin-left: 20px;
+  width: 167.29px;
+  font-weight: 400px;
```
lerastarynets, line 93: The 'font-weight' property doesn't need a 'px' unit. It should only be a number or keyword.

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

src/style.css
Lidia-Pintos-Medina: Taking note. Thank you very much!

PR #3540, opened 2023-08-22

Safina-coder:


src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  top: 25px;
+}
+
+.big-form__img {
+  width: 19px;
+  height: 19px;
+}
+
+.big-form__input,
+.small-form__input {
```
TarasHoliuk, line 54: you did an optimization, but read the checklist point 2 to get know how to do it better (instead of writing `.big-form__input` here and overwriting part of properties for it below)

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  font-family: "Avenir", sans-serif;
```
TarasHoliuk, line 55: you may try `inherit` explicitly to avoid `"Avenir", sans-serif;` duplication but overwrite default browser styles

TarasHoliuk (changes requested):

IvanFesenko (changes requested):
Please fix previous comments.  [Checklist](https://github.com/Safina-coder/layout_search-bar-airbnb/blob/develop/checklist.md)

PR #3578, opened 2023-09-02
Develop

SliwinskiDominik:
https://sliwinskidominik.github.io/layout_search-bar-airbnb/

https://sliwinskidominik.github.io/layout_search-bar-airbnb/report/html_report/


src/style.css
```diff
@@ -0,0 +1,124 @@
+html {
+  --form-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  --form-shadow-hover: 0 3px 8px rgba(61, 78, 97, 0.2);
```
OleziO, line 3: Move all variables to :root selector.

OleziO: Use HEX instead of RGBA

OleziO (changes requested):

plodnikate (approved):

PR #3579, opened 2023-09-02
test is not executed

Miishca:


src/index.html
```diff
@@ -1,16 +1,59 @@
...
+      class="small-form"
+      data-qa="small"
+      method="post"
+      action="/action_page.php"
+    >
+      <div class="small-form__input-container">
+        <div class="small-form__icon">
+          <img class="small-form__img"
+            src="images/Search.svg"
+            alt="Search" >
```
OleziO, line 47:
```suggestion
          <img 
            class="small-form__img"
            src="images/Search.svg"
            alt="Search" 
          >
```

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+}
+
+* {
+  box-sizing: border-box;
+}
+
+html {
+  --font-color: #3d4e61;
+  --white-color: #fff;
+  --border-color: #e1e7ed;
```
OleziO, line 20: Move all variables to the :root selector.

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+
+.big-form__input,
+.small-form__input {
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+  width: 100%;
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  background: #fff;
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
```
OleziO, line 61: Use HEX instead of RGBA.

OleziO: Check other similar cases.

OleziO (changes requested):
Fix view.
image

maxim2310 (approved):
Good work!!!

PR #3640, opened 2023-09-20
Develop

DominoSl:
- [DEMO LINK](https://<your_account>.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://<your_account>.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -59,8 +59,13 @@
   box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
   font-family: inherit;
   font-size: inherit;
-  font-weight: inherit;
   color: inherit;
+  font-weight: 400;
+  text-shadow: 1px 4px 4px rgba(0, 0, 0, 0.25);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,86 @@
...
+  width: 11px;
+  height: 11px;
+  top: 15px;
+  left: 13px;
+}
+
+.form__input {
+  box-sizing: border-box;
+  width: 100%;
+  height: 70px;
```
anastasiia-tilikina, line 54: let's not hard code height for input.
just use proper paddings

anastasiia-tilikina (changes requested):
great job!

let's jus fix one small isssue

src/style.css
DominoSl: Got it! Thanks!

plodnikate (approved):
Awesome!

PR #3657, opened 2023-09-29
search bars

Vasyl1991:
https://Vasyl1991.github.io/layout_search-bar-airbnb/
https://Vasyl1991.github.io/layout_search-bar-airbnb/report/html_report/


VitaliyBodnar (changes requested):
- wrong search field width
Screenshot 2023-09-29 at 13 59 47

PR #3659, opened 2023-09-30
add task solution

VKdrvtsv:
- [DEMO LINK](https://VKdrvtsv.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://VKdrvtsv.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.scss
+++ b/src/style.scss
@@ -1,0 +1,93 @@
+@font-face {
+  font-family: "Avenir";
+  src: url(./fonts/Avenir-Book.ttf), url(./fonts/Avenir-Heavy.ttf);
+}
+
+.form {
+  margin: 0;
+  margin-top: 20px;
+}
+
+.cell {
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  background: #fff;
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  transition: all 0.5s;
+  width: 100%;
+
+  &--big {
+    height: 70px;
+    background-image: url(./images/Search.svg);
+    background-repeat: no-repeat;
+    background-position: left 26px top 25px;
+    background-size: 19px;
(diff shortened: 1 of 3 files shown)


src/style.scss
```diff
@@ -0,0 +1,93 @@
...
+
+.form {
+  margin: 0;
+  margin-top: 20px;
+}
+
+.cell {
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  background: #fff;
```
loralevitska, line 14: It is a good practice to make all colors [variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties). Fix all cases please

src/style.scss
```diff
@@ -0,0 +1,93 @@
...
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  transition: all 0.5s;
+  width: 100%;
+
+  &--big {
+    height: 70px;
+    background-image: url(./images/Search.svg);
+    background-repeat: no-repeat;
+    background-position: left 26px top 25px;
+    background-size: 19px;
```
loralevitska, line 24: Combine with `background` property. Fix all cases

loralevitska (changes requested):
Almost perfect) Let's improve for more practice

src/style.scss
VKdrvtsv: all the background properties make like just background?

loralevitska (changes requested):
Unfortunately i can't review solution due to 404 errors, redeploy and check your links before re-request review

etojeDenys (changes requested):
rewrite your styles following these rule
image
feel free to ask me in chat for some help

VitaliyBondarenko1982 (approved):
Well done

PR #3670, opened 2023-10-05
add task search

AntonBelia:


src/index.html
```diff
@@ -11,6 +11,49 @@
...
+          src="images/Search.svg"
+          alt="search"
+        >
+      </form>
+    </div>
+
+    <div class="search">
+      <form
+        class="search__form"
+        action="http://localhost:8080/"
```
VitaliyBodnar, line 40: - if you want to add some plug for actions is's better to use # instead of hardcoded host 
```suggestion
        action="#"
```

src/index.html
```diff
@@ -11,6 +11,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <div class="search">
+      <form
```
VitaliyBodnar, line 15: don't need two forms here, one form will be enough

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #3680, opened 2023-10-11
add task solution

valost:
https://valost.github.io/layout_search-bar-airbnb/
https://valost.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,82 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 800;
+}
+
+html {
+  --border-color: #e1e7ed;
+  --placeholder-color: #3d4e61;
+  --shadow-color: 0 2px 4px rgba(0, 0, 0, 0.05);
+  --hovershadow-color: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+  --background-color: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
+
+.search {
+  margin-top: 20px;
+}
+
+.input__big {
+  border: 1px solid var(--border-color);
+  border-radius: 4px;
+  box-shadow: var(--shadow-color);
+  width: 100%;
+  background: url(images/Search.svg) no-repeat;
+  background-position: 26px 23px;
+  padding-left: 62px;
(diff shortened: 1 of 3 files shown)


denys-danyliuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,74 @@
...
+}
+
+.input {
+  border: 1px solid var(--border-color);
+  border-radius: 4px;
+  box-shadow: var(--shadow-color);
+  width: 100%;
+  background: var(--background-image);
+  background-size: 19px;
+  background-position: 26px 23px;
```
loralevitska, line 34: `background` shorthand property can be used. Fix all cases

src/style.css
```diff
@@ -0,0 +1,74 @@
...
+  padding-left: 62px;
+  padding-top: 26px;
+  padding-bottom: 26px;
```
loralevitska, line 37: `padding` shorthand property can be used. Fix all cases

src/style.css
```diff
@@ -0,0 +1,74 @@
...
+
+.input:focus {
+  outline-width: 0;
+  font-weight: 800;
+  box-shadow: var(--shadow-color);
+  background: url(images/Search.svg) no-repeat 26px 23px, var(--gradient-color);
+}
+
+.input::placeholder {
+  font-family: Avenir, sans-serif;
```
loralevitska, line 52: Use this property for `html` tag and here only inherit it

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

- try to pass all tests.  Click on each test and switch between `reference` and `test` tabs to recognize difference and fix them
image

valost:
I fixed the code however some tests still don't work.

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.input {
+  border: 1px solid var(--border-color);
+  border-radius: 4px;
+  box-shadow: var(--shadow-color);
+  width: 100%;
+  background: var(--background-image) 26px 23px;
+  background-size: 19px;
+  padding: 27px 62px 28px;
```
Esceype, line 35: Do not hardcode paddings set the height of the form as it on the design, and then change the label display to block and set height 100%, same for input, but you will still need padding-left, and then double-check the top/left/width/height of background and also try to add box-sizing: border-box to `*` selector.

Esceype (changes requested):
GJ! Read the following comment and try to implement it.

maxim2310 (approved):
Good work!!!

PR #3691, opened 2023-10-16
search bar

PonomarchukLana:
- [DEMO LINK](https://ponomarchuklana.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ponomarchuklana.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,28 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form data-qa="big" class="search-form">
+      <div class="search-form__input-wrapp">
+        <span class="search-form__icon search-form__icon--big"></span>
+
+        <input
+          type="search"
+          class="search-form__input search-form__input--big"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,28 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form data-qa="big" class="search-form">
+      <div class="search-form__input-wrapp">
+        <span class="search-form__icon search-form__icon--big"></span>
```
polosanya, line 16: Do you really need extra divs? Better use label instead. 
Also, you can set the background image directly to the input

polosanya (changes requested):
Styles on focus are not implemented
image
image

src/index.html
PonomarchukLana: background-image was redefined when I changed only background on focus(

BudnikOleksii (approved):
Looks good 🔥

PR #3692, opened 2023-10-16
add form solution

yevtok:
- [DEMO LINK](https://yevtok.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://yevtok.github.io/layout_search-bar-airbnb/report/html_report/)

[x] - Add ALL data-qa attributes required in the task
[x] - Text Try "Los Angeles" is a placeholder.
[x] - Use &quot; instead of " in the placeholder.
[x] - Pay close attention to font styles. Use fonts from fonts folder.
[x] - You have to put input inside the form for correct data processing. Form should have action and method attributes.
[x] - Each search bar be inside its own form element.
[x] - Task has styling for :focus state. Forms, labels, divs are not focusable by default.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,14 +3,38 @@
   <head>
     <meta charset="UTF-8">
     <meta
-      name="viewport"
-      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+      name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
     >
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <div class="container">
+      <form
+        class="bigForm"
+        data-qa="big"
+        action="./"
+      >
+        <input
+          type="text"
+          class="input input--big"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -3,14 +3,38 @@
...
+      <form
+        class="bigForm"
+        data-qa="big"
+        action="./"
+      >
+        <input
+          type="text"
+          class="input input--big"
+          data-qa="keypress"
+          placeholder="Try &quot;Los Angeles&quot;"
```
ArtemTeslenko, line 23: class names should be appropriate to BEM, you have parent element - `form`, `form` is a B - bloc in this case, `input` is E - element, `input--big` is M - modifier, so `bigForm`, `input`, `input--big` are not suitable names in this case [BEM](https://mate-academy.github.io/style-guides/htmlcss.html#oocss-and-bem)
the same with smallForm

ArtemTeslenko (changes requested):
Good job! Left comment about BEM, please make sure to go through this topic VERY attentively)

src/style.css
```diff
@@ -0,0 +1,68 @@
...
+  font-style: normal;
+  font-weight: 400;
+  line-height: normal;
+}
+
+.search__input::placeholder {
+  color: #3d4e61;
+}
+
+.search__input--big {
```
polosanya, line 53: Actually, you can make one of them the default one, so another one will be with modifier

polosanya (changes requested):
Styles on focus are absent(
Pay attention to the design
image
image

PR #3701, opened 2023-10-18
search bar almost done

YuriySobakar:
- [DEMO LINK](https://<YuriySobakar>.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://<YuriySobakar>.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -25,12 +25,12 @@ ___
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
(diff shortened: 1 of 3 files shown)


roman-mirzoian (approved):
I see that not all tests pass, but the difference is minimal, so this time it is accepted, but in the future try to make all tests pass locally.
Also, be more careful about the links to the demo, because now your links lead to Google searches.
Well done 👍

PR #3705, opened 2023-10-20
task solution

ksyvvkaa:
- [DEMO LINK](https://ksyvvkaa.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ksyvvkaa.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/styles/blocks/page.css
+++ b/src/styles/blocks/page.css
@@ -1,0 +1,3 @@
+.page__search-bar {
+  margin-top: 20px;
+}
--- a/src/styles/blocks/search-bar.css
+++ b/src/styles/blocks/search-bar.css
@@ -1,0 +1,69 @@
+.search-bar {
+  position: relative;
+}
+
+.search-bar::before {
+  content: "";
+  position: absolute;
+  display: block;
+  background-image: url("./../../images/Search.svg");
+  background-repeat: no-repeat;
+  background-size: contain;
+}
+
+.search-bar--large::before {
+  top: 25px;
+  left: 26px;
+  height: 19px;
+  width: 19px;
+}
+
+.search-bar--small::before {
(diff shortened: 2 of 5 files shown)


src/styles/blocks/page.css
```diff
@@ -0,0 +1,3 @@
+.page__search-bar {
+  margin-top: 20px;
```
ArtemTeslenko, line 2: it is not good practice when margin of the child element is out of parent element, if you need to add space between parent and child borders it is better to use padding for parent, also when you fix it you will not need this class `page__search-bar` any more

src/styles/blocks/search-bar.css
```diff
@@ -0,0 +1,69 @@
...
+
+.search-bar--small::before {
+  top: 15px;
+  left: 13px;
+  height: 11px;
+  width: 11px;
+}
+
+.search-bar__input {
+  box-sizing: border-box;
```
ArtemTeslenko, line 29: move it into reset file

src/styles/blocks/search-bar.css
```diff
@@ -0,0 +1,69 @@
+.search-bar {
+  position: relative;
+}
```
ArtemTeslenko, line 3: you can use nesting here

> .search-bar {
>   &::before {
>   }
>   &--large::before {
>   }
> }

ArtemTeslenko (changes requested):
Great 👍 
Use [nesting](https://mate-academy.github.io/style-guides/htmlcss.html#nested-selectors)
Fix `margin`
Remove extra class
Move common styles into reset file

roman-mirzoian (approved):
Well done 👍

PR #3714, opened 2023-10-23
Develop

Xilston:
- [DEMO LINK](https://xilston.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://xilston.github.io/layout_search-bar-airbnb/report/html_report/)
- [+ ] Icon implemented using background-image CSS property
- [+ ] Inputs are written inside of 'form' tag with correctly passed attributes
- [ +] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ +] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,99 @@
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir.ttc);
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: bold;
+}
+
+.search {
+  display: flex;
+  margin-bottom: 20px;
+}
+
+.search--big {
+  height: 70px;
+  margin-top: 20px;
+}
+
+.search--small {
+  height: 42px;
+  margin-top: 20px;
+}
+
+.search--big .search__icon {
+
+  width: 19px;
+  height: 19px;
+  position: relative;
+  top: 25px;
+  left: 26px;
+  margin-right: -19px;
+}
+
+.search--small .search__icon {
+  background-image: url(images/Search.svg);
+  background-size: contain;
+  background-repeat: no-repeat;
+  width: 11px;
+  height: 11px;
+  position: relative;
+  top: 15px;
+  left: 13px;
+  margin-right: -11px;
+
+}
+
+.search input {
+  border: 1px solid #e1e7ed;
+  width: 100%;
+  border-radius: 4px;
+  background: #fff;
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  font-family: Avenir, sans-serif;
+}
+
+.search input::placeholder {
+  color: #3d4e61;
+}
+
+.search--big input {
+  padding-left: 62px;
+  color: #3d4e61;
+  font-size: 16px;
+  font-style: normal;
+  font-weight: 400;
+  line-height: normal;
+}
+
+.search--small input {
+  padding-left: 33px;
+  color: #3d4e61;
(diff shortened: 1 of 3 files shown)


artemforsoff:
You have a mistake in committing.
I'm sure this happened by accident, but be more careful and check the commits before pushing.
Not _pdeudo_, correct is _pseudo_.

src/style.css
artemforsoff, line 1: [Requirements](https://github.com/mate-academy/layout_search-bar-airbnb#requirements):
* _default` font-weight` must be 300_ - is not implemented

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  background-size: 19px;
+  top: 25px;
+  left: 26px;
+  height: 19px;
+  width: 19px;
+
+}
+
+.search--small::before {
+  content: "";
```
artemforsoff, line 56: Selectors `.search--big::before` and `.search--small::before` have the same rules.
This breaks the rule [Don't repeat yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).
For a solution check [selector lists](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors#selector_lists).

artemforsoff: For selectors `.search--big input` and `.search--small input` same thing.

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+
+.search input::placeholder {
+  color: #3d4e61;
+}
+
+.search--big input {
+  padding-left: 62px;
+  color: #3d4e61;
+  font-size: 16px;
+  font-style: normal;
```
artemforsoff, line 75: Don't copy all properties from Figma, a value of `font-style` is `normal` by default.

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  background: #fff;
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+}
+
+.search input:focus {
+  border-radius: 3px;
+  border: 1px solid #e1e7ed;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
artemforsoff, line 99: If you want to describe one property, it is better not to use a shorthand property.
Because if there were previously declared other properties for the background, this rule clears them.

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  font-size: 14px;
+  font-style: normal;
+  font-weight: 400;
+  line-height: normal;
+}
+
+.search input:hover {
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  background: #fff;
```
artemforsoff, line 92: If you want to describe one property, it is better not to use a shorthand property.
Because if there were previously declared other properties for the background, this rule clears them.

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+}
+
+.search input:focus {
+  border-radius: 3px;
+  border: 1px solid #e1e7ed;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  outline: none;
+  font-weight: bold;
```
artemforsoff, line 101: You've used numeric values in your code before, it's better to keep it that way. This will be useful for global code search.

artemforsoff (changes requested):
There are a couple of mistakes, but overall the job was done good 👌🏻
Look at the comments and correct them.

PR #3729, opened 2023-10-29
add task solution

IvanSolopenko:


src/style.css
```diff
@@ -0,0 +1,123 @@
...
+  font-size: 16px;
+  line-height: 22px;
+}
+
+.form__input_small::placeholder {
+  font-size: 14px;
+  line-height: 22px;
+}
+
+.form__input_big {
```
TarasHoliuk, line 113: you could write it w/o `_big` or `_small` selectors. For example, write all styles for small input (and common styles) inside `.form__input` selector. Then in `.form__input_big` **overwrite** styles you needed (it's described in the checklist).

Anyway, very good job - approved 👍

TarasHoliuk (approved):

PR #3742, opened 2023-11-04
add task solution

Tetiana01:
- [DEMO LINK](https://Tetiana01.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Tetiana01.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,31 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="search-bar search-bar--max" data-qa="big">
+      <div class="search-bar__icon search-bar__icon--max">
+      </div>
+
+      <input
+        class="search-bar__input search-bar__input--max"
+        type="text"
+        name="search1"
+        placeholder="Try “Los Angeles”"
+        id="search1"
+        data-qa="keypress"
+      >
+    </form>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,31 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="search-bar search-bar--max" data-qa="big">
+      <div class="search-bar__icon search-bar__icon--max">
+      </div>
```
etojeDenys, line 16: better to use img tag instead of the div with background-image

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  font-weight: 400;
+}
+
+@font-face {
+  font-family: "Avenir-Heavy";
+  src: url("./fonts/Avenir-Heavy.ttf");
+  font-style: normal;
+  font-weight: 400;
+}
+
```
etojeDenys, line 14: why do you need two different fonts for the same font-style and font-weight

etojeDenys (changes requested):
let's stretch the input
image

src/index.html
Tetiana01: In the requirements we have:
Знімок екрана 2023-11-07 001731

OleziO (approved):

PR #3746, opened 2023-11-06
add search-bar solution

BohdanDymydiuk:
- [DEMO LINK](https://BohdanDymydiuk.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://BohdanDymydiuk.github.io/layout_search-bar-airbnb/report/html_report/)
- [x] All Typical Mistakes from BEM lesson theory are checked.
- [x] Code follows all the [Code Style Rules ❗️](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -8,12 +8,17 @@
   src: url("fonts/Avenir-Heavy.ttf");
 }
 
+html {
+  --text-color: #3d4e61;
+}
+
 * {
   box-sizing: border-box;
 }
 
 body {
   font-weight: 300;
+  font-family: "Avenir Book", sans-serif;
 }
 
 .search-bar {
@@ -48,8 +53,8 @@ body {
 }
 
 .input-field {
-  font-family: "Avenir Book", sans-serif;
-  color: #3d4e61;
+  font-family: inherit;
+  color: var(--text-color);
   border: none;
   outline: none;
   padding: 0;
@@ -69,7 +74,7 @@ body {
 }
 
 ::placeholder {
-  color: #3d4e61;
+  color: var(--text-color);
 }
 
 .input-field--focused:focus {
--- a/src/style.css
+++ b/src/style.css
@@ -77,11 +77,7 @@
   color: #3d4e61;
 }
 
-.search-bar--large__input-field:focus {
-  font-family: "Avenir Heavy", sans-serif;
-}
-
-.search-bar--small__input-field:focus {
+.input-field--focused:focus {
   font-family: "Avenir Heavy", sans-serif;
 }
 
--- a/src/style.css
+++ b/src/style.css
@@ -9,47 +9,42 @@
 }
 
 * {
-  font-weight: 300;
   box-sizing: border-box;
+}
+
+body {
+  font-weight: 300;
 }
 
 .search-bar {
   display: flex;
   align-items: center;
   margin-top: 20px;
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+}
+
+.search-bar:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
 }
 
 .search-bar--large {
   height: 70px;
   padding-left: 26px;
-  border-radius: 4px;
-  border: 1px solid #e1e7ed;
-  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
 }
 
 .search-bar--large:hover {
   box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.25);
 }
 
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,113 @@
...
+  font-family: "Avenir Heavy";
+  src: url("fonts/Avenir-Heavy.ttf");
+}
+
+body {
+  font-weight: 300;
+}
+
+.largeSearchBar,
+.smallSearchBar {
```
IvanFesenko, line 16: use kebab-case for class naming. Also use BEM methodology

src/style.css
```diff
@@ -0,0 +1,113 @@
...
+  display: flex;
+  align-items: center;
+  margin-top: 20px;
+  box-sizing: border-box;
```
IvanFesenko, line 20: Move `box-sizing: border-box;` into `*` selector

IvanFesenko (changes requested):
Please check [checklist](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md) before send to next review

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  color: #3d4e61;
+}
+
+.search-bar--large__input-field:focus {
+  font-family: "Avenir Heavy", sans-serif;
+}
+
+.search-bar--small__input-field:focus {
+  font-family: "Avenir Heavy", sans-serif;
+}
```
anastasiia-tilikina, line 86: seems like this code is duplicated.
why do you set font-family on focus but not by default?

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
+
+.search-bar--small {
+  height: 42px;
+  padding-left: 13px;
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+}
```
anastasiia-tilikina, line 44: The code for small and big search bars is duplicated.
It would be better to set styles for search-bar class and override for small one only those you need. for instance:

```css
.search-bar {
  height: 70px;
  padding-left: 26px;
  border-radius: 4px;
  border: 1px solid #e1e7ed;
  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
}

.search-bar--small {
  height: 42px;
  padding-left: 13px;
}
``` 

with this approach you write less code and don't duplicate it)

anastasiia-tilikina: fix everywhere

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  src: url("fonts/Avenir-Book.ttf");
+}
+
+@font-face {
+  font-family: "Avenir Heavy";
+  src: url("fonts/Avenir-Heavy.ttf");
+}
+
+* {
+  font-weight: 300;
```
anastasiia-tilikina, line 12: keep in * selector only for seting box-sizing property. do not set other properties with this selector to avoid lack of perfomance

anastasiia-tilikina (changes requested):
check my comments

src/style.css
BohdanDymydiuk: because that's the font-family that should be in focus according to the layout

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+  height: 42px;
+  padding-left: 13px;
+}
+
+.search-bar--small:hover {
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+}
+
+.input-field {
+  font-family: "Avenir Book", sans-serif;
```
loralevitska, line 51: Move to body selector
and here just inherit it

loralevitska (approved):
Good job!
Check my comments for improvement

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+  font-family: "Avenir Heavy", sans-serif;
+}
+
+.button-icon {
+  border: none;
+  background-color: #fff;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: contain;
```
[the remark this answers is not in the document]
BohdanDymydiuk: I tried it. The icon did not appear.

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+  color: #3d4e61;
+}
+
+.input-field--focused:focus {
+  font-family: "Avenir Heavy", sans-serif;
+}
+
+.button-icon {
+  border: none;
+  background-color: #fff;
```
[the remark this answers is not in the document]
BohdanDymydiuk: if I do this, the background will be gray, not like in the layout

PR #3747, opened 2023-11-06
add task solution

Mariana-VV:
[x ] Icon implemented using background-image CSS property
- [x ] Inputs are written inside of 'form' tag with correctly passed attributes
- [ x] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ x] Code follows all the [Code Style Rules ❗️](./checklist.md)


Esceype (changes requested):
Good job, read the following comments below.

PR #3753, opened 2023-11-07
search-bar

StepanenkoA:
- [DEMO LINK](https://StepanenkoA.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://StepanenkoA.github.io/layout_search-bar-airbnb/report/html_report/)


src/index.html
```diff
@@ -11,6 +11,22 @@
...
+
+      <div data-qa="small"
+      class="form__search form__search--small">
+        <div class="form___icon form__icon--small"></div>
+        <input class="form__input form__input--small"
+        placeholder="Try “Los Angeles”">
+      </div>
+    </form>
+
+
```
anastasiia-tilikina, line 30:
```suggestion
```
do not leave epmty lines between children and parent tags

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+    url(/fonts/Avenir.ttc);
+  font-weight: 900;
+}
+
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
```
anastasiia-tilikina, line 14: don't set these properties via * selector.
use * only for setting box-sizing property

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+.form {
+  display: flex;
+  flex-direction: column;
+  padding-left: 8px;
+  padding-top: 20px;
+  padding-right: 8px;
```
anastasiia-tilikina, line 22: combine this into one property

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+.form__search--big {
+  padding-top: 23px;
+  padding-bottom: 23px;
```
anastasiia-tilikina, line 27: the same

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+.form__input--big::placeholder {
+  color: rgba(61, 78, 97, 1);
+  font-weight: 300;
+}
+
+.form__search--small {
+  padding-top: 10px;
+  padding-bottom: 10px;
```
anastasiia-tilikina, line 41: same here

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+.form__search--small:focus {
+  font-family: "Avenir", sans-serif;
```
anastasiia-tilikina, line 45:
```suggestion
  font-family: inherit;
```
it's better to use inherit font family

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+.form__icon--small {
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: cover;
+  display: inline-block;
+  width: 11px;
+  height: 11px;
+  margin-right: 9px;
+  margin-left: 13px;
+}
```
anastasiia-tilikina, line 95: you're duplicating your code here.
use common class to set properties and override for small one those you need.

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+  font-size: 16px;
+  border: 0;
+  outline: none;
+}
+
+.form__input--small {
+  font-size: 14px;
+  border: 0;
+  outline: none;
+}
```
anastasiia-tilikina, line 111: the same here.

anastasiia-tilikina: check everywhere.

src/style.css
anastasiia-tilikina: for example:

```css
.form__icon {
// styles for big icon
}

.fom_icon--small {
// styles that are different from bigger one
}
```

anastasiia-tilikina (changes requested):
check my coments

src/index.html
```diff
@@ -11,6 +11,20 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="form">
+      <div data-qa="big" class="form__search form__search--big">
+        <div class="form___icon form__icon--big"></div>
+        <input class="form__input form__input--big"
+          placeholder="Try “Los Angeles”"
+          data-qa="keypress">
```
VitaliyBodnar, line 19: code is not aligned, please check other places 
```suggestion
        <input 
          class="form__input form__input--big"
          placeholder="Try “Los Angeles”"
          data-qa="keypress"
         >
```

src/style.css
```diff
@@ -0,0 +1,103 @@
...
+  font-weight: 900;
+}
+
+body {
+  margin: 0;
+}
+
+* {
+  box-sizing: border-box;
+  padding: 0;
```
VitaliyBodnar, line 15: only box-sizing can be in this block `*{}`

src/style.css
```diff
@@ -0,0 +1,103 @@
...
+  flex-direction: column;
+  padding: 20px 8px 8px;
+}
+
+.form__search--big {
+  padding: 23px 0;
+}
+
+.form__search--big:hover {
+  box-shadow: 0 4px 4px 0 rgba(0,0,0, 0.25);
```
VitaliyBodnar, line 29: It is a good practice to make all colors [variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

src/style.css
```diff
@@ -0,0 +1,103 @@
...
+  font-weight: 900;
+}
+
+.form__search--small:hover {
+  box-shadow: 0 3px 8px 0 rgba(61,78,97, 0.2);
+}
+
+.form__input--small::placeholder {
+  color: rgba(61, 78, 97, 1);
+  font-weight: 300;
```
VitaliyBodnar, line 52: font-weight: 300 is duplicated, so make sense to move this style to form__input

src/style.css
```diff
@@ -0,0 +1,103 @@
...
+.form___icon {
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: cover;
+  display: inline-block;
+  width: 19px;
+  height: 19px;
+  margin-left: 26px;
+  margin-right: 17px;
+  flex-shrink: 0;
```
VitaliyBodnar, line 80: Let's try to organize your CSS
I recommend you take a look at these styles code and use it in the feature
[9elements](https://9elements.com/css-rule-order) or [htmlAcademy](https://codeguide.academy/html-css.html#css-order)

VitaliyBodnar (changes requested):

volodymyr-soltys97 (changes requested):
The input field should be the full width of the screen, we also need to check the styles for the font when we enter the text as on the design
image

xiimiik (approved):
nice work!

PR #3769, opened 2023-11-12
add task solution

sophiasph:
- [DEMO LINK](https://sophiasph.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://sophiasph.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,24 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <div class="search-bar">
+      <label for="search-input__big"></label>
+      <input
+        type="text"
+        id="search-input__big"
+        class="search-bar__big"
+        data-qa="big keypress"
+        placeholder="Try &quot;Los Angeles&quot;"
+      >
+
+      <label for="search-input__small"></label>
+      <input
+        type="text"
+        id="search-input__small"
+        class="search-bar__small"
+        data-qa="small"
+        placeholder="Los Angeles"
+      >
+    </div>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,79 @@
+@font-face {
+  font-family: "Avenir Book";
+  src: url("./fonts/Avenir-Book.ttf");
+  font-weight: normal;
+  font-style: normal;
+}
+
+@font-face {
+  font-family: "Avenir Heavy";
+  src: url("/fonts/Avenir-Heavy.ttf");
+  font-weight: normal;
+  font-style: normal;
+}
+
+:root {
+  --text-color: rgb(61, 78, 97);
+  --bg-gradi: linear-gradient(180deg, rgb(255, 255, 255), rgb(246, 246, 247));
+}
+
+body {
+  margin: 0;
+  font-weight: 300;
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  background: url("./images/Search.svg") 26px 25px no-repeat, var(--bg-gradi);
+}
+
+.search-bar__big:focus::placeholder {
+  text-shadow: 1px 4px 4px rgba(0, 0, 0, 0.25);
+  font-weight: 800;
+}
+
+.search-bar__small {
+  border: 1px solid #e1e7ed;
```
TarasHoliuk, line 62: you repeat a lot of values that are common for both - big and small

do the following:

```

.search-bar {
  border: 1px solid #e1e7ed;
  margin-top: 20px;
  width: 100%;
  max-width: 900px;
  padding: 25px 62px;
  background: url("./images/Search.svg") 26px 25px no-repeat;
}

.search-bar__small {
  set only values that not the same as in big input
}
```
in HTML for big input use `search-bar` class only, for small input - both `search-bar search-bar-small`

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: 20px;
+  width: 100%;
+}
+
+.search-bar__big {
+  border: 1px solid #e1e7ed;
+  margin-top: 20px;
+  width: 100%;
```
TarasHoliuk, line 40: it doesn't work
image

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  max-width: 900px;
+  padding: 25px 62px;
+  background: url("./images/Search.svg") 26px 25px no-repeat;
+}
+
+.search-bar__big:hover {
+  box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.25);
```
TarasHoliuk, line 47: w/o `hover` it also have shadow, but a bit different - check the design

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  font-family: "Avenir Book";
+  src: url("./fonts/Avenir-Book.ttf");
+  font-weight: normal;
+  font-style: normal;
+}
+
+@font-face {
+  font-family: "Avenir Heavy";
+  src: url("/fonts/Avenir-Heavy.ttf");
+  font-weight: normal;
```
TarasHoliuk, line 11: it's bold, not normal

TarasHoliuk (changes requested):

lerastarynets (changes requested):
well done in fixing comments )
Pls try to pass all tests from backstop before asking for review again)

src/index.html
```diff
@@ -11,6 +11,24 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <div class="search-bar">
```
volodymyr-soltys97, line 14: Use a semantic tag form instead of div

src/style.css
volodymyr-soltys97: Still not fixed

volodymyr-soltys97 (changes requested):
1. Need to remove the default border when input is focusing
image

2. Need to fix the problem with all tests
you need to check the styles for the text, also the inputs should be the full width of the browser window

PR #3779, opened 2023-11-15
add task solution

prokopenkooleh:
- [DEMO LINK](https://prokopenkooleh.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://prokopenkooleh.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,91 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 400;
+}
+
+@font-face {
+  font-family: Avenir;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,91 @@
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 400;
```
volodymyr-soltys97, line 4: Default value, you can just remove it
```suggestion
```

volodymyr-soltys97 (approved):
Nice work 🔥

src/style.css
prokopenkooleh: Ok, thank you

PR #3783, opened 2023-11-16
Develop

tndeineka:
[DEMO LINK](https://tndeineka.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://tndeineka.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,73 @@
+@font-face {
+  font-family: "Avenir";
+  font-style: normal;
+  font-weight: 300;
+  src: url("./fonts/Avenir-Book.ttf");
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.form {
+  display: flex;
+  width: 100%;
+  height: 70px;
+  margin-top: 20px;
+}
+
+.label {
+  display: block;
+  position: relative;
+  width: 100%;
+}
+
+.label::before {
+  content: "";
+  position: absolute;
+  background-image: url("images/Search.svg");
+  background-size: contain;
+  background-repeat: no-repeat;
+  width: 19px;
+  height: 19px;
+  top: 25px;
+  left: 26px;
+  z-index: 1;
+}
+
+.label--small::before {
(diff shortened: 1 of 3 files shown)


Andriy-Shymkiv (changes requested):
test link is not working, please fix it and request a review again

src/style.css
```diff
@@ -0,0 +1,74 @@
+@font-face {
+  font-family: "Avenir";
+  font-style: normal;
+  font-weight: 300;
+  src: url("./fonts/Avenir-Book.ttf");
+}
+
+* {
+  font-family: "Avenir", sans-serif;
+  box-sizing: border-box;
```
loralevitska, line 10: Only `box-sizing` property allowed for `*` selector. `font-family` better move to body or html tag

src/style.css
```diff
@@ -0,0 +1,74 @@
...
+  position: relative;
+  width: 100%;
+}
+
+.label::before {
+  content: "";
+  position: absolute;
+  background-image: url("images/Search.svg");
+  background-size: contain;
+  background-repeat: no-repeat;
```
loralevitska, line 31: Can be combined with [`background` shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/background)

src/style.css
```diff
@@ -0,0 +1,74 @@
...
+}
+
+.search {
+  width: 100%;
+  height: 70px;
+  padding-left: 62px;
+  font-family: inherit;
+  font-size: 16px;
+  font-weight: 400;
+  color: #3d4e61;
```
loralevitska, line 53: It is a good practice to make all colors [variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

loralevitska (changes requested):
Almost perfect) Let's improve for more practice

xiimiik (approved):
nice work!

PR #3785, opened 2023-11-18
add task solution

SofiiaYarymovych:
-[DEMO LINK](https://SofiiaYarymovych.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://SofiiaYarymovych.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,78 @@
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir.ttc);
+}
+
+/* @font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Heavy.ttf);
+} */
+
+.search__bar {
+  box-sizing: border-box;
+  display: flex;
+  align-items: center;
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  background: #fff;
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  margin-top: 20px;
+}
+
+.search__icon {
+  background-image: url(images/Search.svg);
+  width: 19px;
+  height: 19px;
+  flex-shrink: 0;
+  display: block;
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: contain;
+  margin-left: 26px;
+}
+
+.search__input {
+  width: 100%;
+  border: none;
+  color: #3d4e61;
+  font-family: "Avenir", sans-serif;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,78 @@
...
+
+/* @font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Heavy.ttf);
+} */
```
loralevitska, line 14: Remove comment. Try it write obvious code that explains all we need without comments)

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+
+.search__bar {
+  box-sizing: border-box;
+  display: flex;
+  align-items: center;
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
```
loralevitska, line 21: It is a good practice to make all colors [variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+
+.search__icon {
+  background-image: url(images/Search.svg);
+  width: 19px;
+  height: 19px;
+  flex-shrink: 0;
+  display: block;
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: contain;
```
loralevitska, line 35: Combine it with [`background`](https://developer.mozilla.org/en-US/docs/Web/CSS/background) shorthand property

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.search__input {
+  width: 100%;
+  border: none;
+  color: #3d4e61;
+  font-family: "Avenir", sans-serif;
+  font-size: 16px;
+  font-weight: 300;
+  font-style: normal;
```
loralevitska, line 46: Remove redundant default styles

loralevitska (changes requested):
Almost done) Let's improve your solution 

- switch between `reference` and `test` tabs to recognize difference and fix them (pay attention to small search bar height, font-family and inner indents)
image

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  --text-color: #3d4e61;
+}
+
+.body {
+  font-family: "Avenir", sans-serif;
+  font-weight: 300;
+}
+
+.search__bar {
+  box-sizing: border-box;
```
VitaliyBondarenko1982, line 18: general practice to set box-sizing property  in * selector.

VitaliyBondarenko1982 (approved):
Looks good.

PR #3808, opened 2023-11-25
Added solution

Diliradon:
Щось не розумію добавив але не те.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -3,6 +3,16 @@
   src: url(fonts/Avenir.ttc);
 }
 
+@font-face {
+  font-family: Avenir, sans-serif;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+@font-face {
+  font-family: Avenir, sans-serif;
+  src: url(fonts/Avenir-Book.ttf);
+}
+
 * {
   box-sizing: border-box;
 }
@@ -14,11 +24,12 @@
   width: 100%;
   border-radius: 4px;
   border: 1px solid #e1e7ed;
+  background: #fff;
   box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
 }
 
 .form__search {
-  font-family: "Avenir", sans-serif;
+  font-family: "Avenir-Book", sans-serif;
   font-weight: 300;
   display: flex;
   align-items: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  border: var(--border-sol-col);
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+}
+
+.search__input:focus {
+  width: 100%;
+  outline: none;
+  border-radius: 3px;
+  border: var(--border-sol-col);
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
VitaliyBodnar, line 84: Color #fff is duplicated, so make sense to move it to the variable 
Check other duplicated styles

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+}
+
+.search__input {
+  width: 100%;
+  padding-left: 62px;
+  border-radius: 3px;
+  border: none;
+  height: 100%;
+  font-size: 16px;
+  font-weight: 400;
```
VitaliyBodnar, line 69: 400 is a default font-weight, so it's redundant

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+* {
+  box-sizing: border-box;
+}
+
+:root {
+  --border-sol-col: 1px solid #e1e7ed;
+}
+
+.form {
+  font-family: "Avenir", sans-serif;
```
VitaliyBodnar, line 25: It's better to font-family to tag body as it will make you code more scalable

VitaliyBodnar (changes requested):

VitaliyBodnar (changes requested):
Please ask you question in a fe_chat

src/style.css
Diliradon: but #fff  its color for linear-gradient  in input. another background-color: #fff  its for Form

Diliradon: I think It`ll be little mix up

VitaliyBodnar (changes requested):
- All tests are not passed. If you have some issues with it please ask it in a fe_chat.

image

Diliradon:
We tried to resolve this problem with a mentor but were unsuccess. We couldn't solve this problem. There one two different pixels.

etojeDenys (approved):
good job

PR #3809, opened 2023-11-25
add search bars

manch0ffline:
- [DEMO LINK](https://manch0ffline.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://manch0ffline.github.io/layout_search-bar-airbnb/report/html_report/)


src/index.html
```diff
@@ -11,6 +11,24 @@
...
   <body>
-    <h1>Search bar airbnb</h1>
+
+    <div class="search search--big" data-qa="big">
+      <div class="search--bigimage"></div>
+      <input type="text"
+      class="key key--big"
+      placeholder="Try “Los Angeles”"
+      data-qa="keypress"
+      >
```
loralevitska, line 21: Fix all cases

```suggestion
      <input 
        type="text"
        class="key key--big"
        placeholder="Try “Los Angeles”"
        data-qa="keypress"
      >
```

src/style.css
```diff
@@ -1 +1,96 @@
 /* add styles here */
+@font-face {
+  font-family: Avenir;
+  src:
+  local("Avenir"),
+  url(fonts/Avenir.ttc),
+  url(fonts/Avenir-Book.ttf),
+  url(fonts/Avenir-Heavy.ttf);
```
loralevitska, line 8: Better use 3 font-faces with different font-weight value

src/style.css
```diff
@@ -1 +1,96 @@
...
+}
+
+.search {
+  display: flex;
+  border-radius: 4px;
+  border: 1px solid #E1E7ED;
```
loralevitska, line 14: It is a good practice to make all colors [variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

src/style.css
```diff
@@ -1 +1,96 @@
...
+
+.search:focus {
+  border-radius: 3px;
+  outline: none;
+  background: linear-gradient(180deg, #FFF 0%, #F6F6F7 100%);
+}
+
+.search--bigimage {
+  margin-left: 26px;
+  margin-right: 17px;
```
loralevitska, line 48: Combine with `margin` property. Fix all cases

src/style.css
```diff
@@ -1 +1,96 @@
...
+  transition: box-shadow 0.3s ease;
+}
+
+.key {
+  outline: none;
+  border: none;
+  padding: 0;
+  font-weight: 300;
+  color: #3D4E61;
+  font-family: Avenir, sans-serif;
```
loralevitska, line 28: Move to html selector and there only inherit this property. Fix all cases if exist

loralevitska (changes requested):
Please redeploy your solution for review
image

PR #3820, opened 2023-11-28
add task solution

styopochkin23:


src/style.css
```diff
@@ -0,0 +1,75 @@
...
+  background-repeat: no-repeat;
+  background-position: 13px 15px;
+  padding-left: 33px;
+  font-size: 14px;
+}
+
+input::placeholder {
+  font-weight: 300;
+  color: #3d4e61;
+}
```
sTorba24, line 75: not best idea to use tags to style
use classes instead

sTorba24 (changes requested):

sTorba24 (changes requested):
Well done! Lets organize it perfectly

src/style.css
```diff
@@ -0,0 +1,75 @@
...
```
sTorba24, line 75: still used tags

u can create class input and put in on both inputs

style class input as you need for big 

then add class `input-small` to small and rewrite some styles that you need

sTorba24 (changes requested):

Esceype (approved):
Well done!

PR #3829, opened 2023-11-30
add

avakiel:
- [DEMO LINK](https://avakiel.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://avakiel.github.io/layout_search-bar-airbnb/report/html_report/)


src/index.html
```diff
@@ -1,16 +1,41 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="UTF-8">
     <meta
       name="viewport"
-      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
-    >
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
```
olya-shyian, line 7: this symbol `>` which is now in the end of the tag should be on the next new line

```
<meta
  name="viewport"
  content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
>
```

src/index.html
```diff
@@ -1,16 +1,41 @@
...
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="form"
+      data-qa="big"
+    >
```
olya-shyian, line 15: please fix attributes formatting: each attribute shold be in the new line

image

[Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

src/index.html
```diff
@@ -1,16 +1,41 @@
...
-      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
-    >
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Search bar airbnb</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="form"
```
olya-shyian, line 13: form tag also should have the next attributes

 action="./"
 method="get"

src/index.html
```diff
@@ -1,16 +1,41 @@
...
+          name="form__input"
+          data-qa="keypress"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
+
+    <form class="form form_small"
+      data-qa="small"
+    >
```
olya-shyian, line 29: please fix attributes formatting

src/style.css
```diff
@@ -1 +1,105 @@
...
+/* stylelint-disable font-family-name-quotes */
+
+* {
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: "Avenir";
+  font-weight: 300;
+  font-style: normal;
```
olya-shyian, line 10: there is no need to write font-style: normal; cause it works by default

please delete this rule in other places as well

src/index.html
```diff
@@ -1,16 +1,41 @@
...
+          class="form__input"
+          type="text"
+          name="form__input"
+          data-qa="keypress"
+          placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </label>
+    </form>
+
+    <form class="form form_small"
```
olya-shyian, line 27: according to BEM if you want to create a modifier your class should be "form--small"

check in other places, and instead of `_` use `--`

src/style.css
```diff
@@ -1 +1,105 @@
...
+
+body {
+  font-family: "Avenir", sans-serif;
+}
+
+.form {
+  display: flex;
+  justify-content: center;
+  margin: 20px auto 0;
+  width: 100%;
```
olya-shyian, line 29: there is no need to write width on your parent block cause this block will stretch by child elements

also you should try to write the BEM block (in our case form) in such a way that your block will not be broken if you move it to another place.

therefore, ideally, the sizes and positioning of blocks should not be explicitly specified. it is better to explicitly set the width-height to the child elements (in our case input elements) and form block are stretched automatically

src/style.css
```diff
@@ -1 +1,105 @@
-/* add styles here */
+/* stylelint-disable font-family-name-quotes */
+
+* {
+  box-sizing: border-box;
+}
```
olya-shyian, line 5: let's move this style block after @font-face;
so that at first we connected the fonts and then there was all the information about the styles

olya-shyian (changes requested):
Terrific work! ✨

src/index.html
avakiel: mb I don't understand the formatting principle very well... but ok, I'll fix it

package.json
```diff
@@ -22,8 +22,8 @@
     "@mate-academy/bemlint": "^0.1.1",
     "@mate-academy/eslint-config": "*",
     "@mate-academy/linthtml-config": "*",
+    "@mate-academy/scripts": "^1.2.12",
```
SerhiiKirik, line 25: In the future - Do not add extra files to PR
To add the file to a commit use `git add src/.` & `git commit -m 'your message'`

SerhiiKirik (approved):
Well done 🔥 
Let's check comments 🌎

PR #3849, opened 2023-12-06
add task solution

ilyaivasyk:
- [DEMO LINK](https://ilyaivasyk.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ilyaivasyk.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -103,3 +103,5 @@ input {
   left: 13px;
   transform: translateY(-50%);
 }
+
+/* finish */


src/style.css
```diff
@@ -1 +1,98 @@
...
+  padding: 23px 60px;
+  font-size: 16px;
+}
+
+.form__form-sm__input {
+  padding: 11px 31px;
+  font-size: 14px;
+}
+
+.form__form-big__input:focus, .form__form-sm__input:focus {
```
Esceype, line 61: Create one general form input class you don`t need to create two classes

src/style.css
```diff
@@ -1 +1,98 @@
...
+
+.form__form-big--icon::before {
+  position: absolute;
+  content: '';
+  background: url('images/Search.svg') center;
+  background-size: cover;
+  width: 19px;
+  height: 19px;
+  top: 50%;
+  left: 25px;
```
Esceype, line 84: Looks like this is wrong values please double check the design

src/style.css
```diff
@@ -1 +1,98 @@
...
+
+.form__form-big__input, .form__form-sm__input {
+  width: 100%;
+  border: var(--border);
+  box-shadow: var(--box-shadow-input);
+  border-radius: 4px;
+}
+
+.form__form-big__input {
+  padding: 23px 60px;
```
Esceype, line 52: This padding cut the image as you can see and that's one of the problems keeping your tests from being successful.

Esceype (changes requested):
GJ!
Read the following comments and please double check the values in design

ilyaivasyk:
Меніне дали відповідь на питання в чаті(

l4st1m0za (approved):
Great job! I can see that you have fixed the issue, but for the next tasks, please, don't forget to ran the tests so that they will be based on your latest changes

PR #3851, opened 2023-12-06
layout_search-bar-airbnb

DmitryVolchokFE:
[DEMO LINK](https://DmitryVolchokFE.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://DmitryVolchokFE.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,24 +1,24 @@
 @font-face {
   font-family: Avenir;
-  font-style: normal;
   font-weight: 300;
   src: url(./fonts/Avenir-Book.ttf);
 }
 
 @font-face {
   font-family: Avenir;
-  font-style: normal;
   font-weight: 800;
   src: url(./fonts/Avenir-Heavy.ttf);
 }
 
-html {
-  font-weight: 300;
-
+:root {
   --lite-gray: #e1e7ed;
   --wite: #fff;
   --bg-color: #f6f6f7;
   --text-color: #3D4E61;
+}
+
+html {
+  font-weight: 300;
 }
 
 .main {
@@ -44,12 +44,12 @@ html {
 
 .search--big {
   height: 70px;
-  padding-left: 25px;
+  padding-left: 25.5px;
 }
 
 .search--small {
   height: 42px;
-  padding-left: 11px;
+  padding-left: 12px;
 }
 
 .search:hover {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,102 @@
 /* add styles here */
```
OleziO, line 1: Remove all comments from the code.

src/style.css
```diff
@@ -1 +1,102 @@
...
+  gap: 20px;
+  flex-direction: column;
+}
+
+.search {
+  display: flex;
+  align-items: center;
+  border-radius: 4px;
+  border: 1px solid #E1E7ED;
+  background-color: #FFF;
```
OleziO, line 33: Move all repeatable colors to the CSS variables.

OleziO (changes requested):
Make input full width and height
image

src/style.css
```diff
@@ -1 +1,116 @@
...
+  width: 100%;
+}
+
+.search__input:focus, .search__input:active {
+  outline: 0;
+}
+
+.search__input--big {
+  font-size: 16px;
+  font-style: normal;
```
TarasHoliuk, line 102: it's default value - no need to set it
FYI

src/style.css
```diff
@@ -1 +1,116 @@
...
+.search--small {
+  height: 42px;
+  padding-left: 11px;
+}
+
+.search:hover {
+  box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.25);
+}
+
+.search:active {
```
TarasHoliuk, line 55: not `active`, but `:focus`
read about these states of an element: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus

Also, now it applies only to the small area on the left side
Screenshot from 2023-12-07 20-09-23

you should add gradient to input on focus

src/style.css
```diff
@@ -1 +1,116 @@
...
+  width: 11px;
+  height: 11px;
+  position: relative;
+  top: -1px;
+}
+
+.search__input {
+  display: inline-block;
+  border: none;
+  color: var(--text-color);
```
TarasHoliuk, line 89: set also font-weight here - to add style for typed text

TarasHoliuk (changes requested):
Well done :+1: 
Let's improve it a bit - I've added a few comments

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #3857, opened 2023-12-08
Develop

nila20:
- [DEMO LINK](https://nila20.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://nila20.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,9 +1,13 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Document</title>
+    <meta
+      name="viewport"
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -1 +1,114 @@
...
+.search-form__label--big {
+  width: 100%;
+  height: 70px;
+}
+
+.search-form__label--small {
+  height: 42px;
+}
+
+.search-form__input {
```
TarasHoliuk, line 42: should also have bold (600 in your case) font-weight - use `Avenir-Heavy.ttf` for typed text

but since everything else LGTM - approved (no need to re-request review) - this comment only FYI

nila20: Im not quite understand, how to use Avenir-Heavy.ttf. Can you explain it more please?

PR #3916, opened 2023-12-29
solution to the task SEARCH BARS

ashatskyy:
- [DEMO LINK](https://ashatskyy.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ashatskyy.github.io/layout_search-bar-airbnb/report/html_report/)

 - [x] All Typical Mistakes from BEM lesson theory are checked.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,134 @@
-/* add styles here */
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  src: url("./fonts/Avenir-Book.ttf");
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 600;
+  src: url("./fonts/Avenir-Heavy.ttf");
+}
+
+body {
+  font-family: Avenir, sans-serif;
+}
+
+.big-form {
+  display: flex;
+	margin-top: 20px;
+	width: 100%;
+  height: 70px;
+}
+
+.big-form__label {
+	display: block;
+  position: relative;
+  width: 100%;
+}
+
+.big-form__label::after {
+  display: block;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,134 @@
...
+}
+
+.big-form__input {
+	box-sizing: border-box;
+  font-weight: 300;
+	font-size: 16px;
+	line-height: 22px;
+  width: 100%;
+  height: 100%;
+  border: 1px solid #e1e7ed;
```
anastasiiavorobiova, line 50: Consider using variables for repeating colors

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/style.css
```diff
@@ -1 +1,137 @@
...
+	background-size: contain;
+	width: 11px;
+	height: 11px;
+	top: 15px;
+	left: 13px;
+}
+
+.small-form__input {
+	box-sizing: border-box;
+  font-weight: 300;
```
TarasHoliuk, line 107: You are repeating a lot of styles for small and big forms.

The way to do it correctly, w/o repeating yourself:
- Create a common class `form__input` and write there all the styles that are needed (for big input for example).
- Create `form__input--small` class and **overwrite** only those properties that are different in small input.
- For small input in HTML use both classes: common `form__input` , and modificator `form__input--small`

That's related to all elements

TarasHoliuk (approved):
Everything looks good, you've fixed all the previous comments so approved :heavy_check_mark: 

But added 1 important comment FYI - read it please

PR #3930, opened 2024-01-04
solution

Maryna1313:
- [DEMO LINK](https://maryna1313.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://maryna1313.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,87 @@
-/* add styles here */
+@font-face {
+  font-family: Avenir;
+  font-style: normal;
+  font-weight: 300;
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir;
+  font-style: normal;
+  font-weight: 800;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+:root {
+  --font-color: #3D4E61;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,87 @@
...
+
+.form__label {
+  display: flex;
+  width: 100%;
+}
+
+.form__input {
+  width: 100%;
+  color: var(--font-color);
+  font-weight: 400;
```
l4st1m0za, line 40: Default font-weight must be 300, as specified in the task description. That may be the reason why your tests are broken.

l4st1m0za (changes requested):
Good job, but it's better to try to pass all the tests

anastasiia-tilikina (changes requested):
your font-weight is still wrong and inputs have wrong font-family
image

remember that inputs and placeholders don't inherit font-family by default. we need to set it explicitly. use `font-family: inherit`

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

1.still have wrong font and font weight 
image
2. try to pass all tests

denys-danyliuk (approved):

PR #3939, opened 2024-01-07
create forms

AnnSomko:


src/index.html
```diff
@@ -11,6 +11,26 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <form class="form form-big"
+          data-qa="big"
+          action="/">
```
TarasHoliuk, line 16: Check the code-style rules: [checklist](https://github.com/mate-academy/layout_search-bar-airbnb/blob/master/checklist.md)

image

src/index.html
```diff
@@ -11,6 +11,26 @@
...
+      <img class="form__img-big"
+           src="images/Search.svg"
+           alt="search">
+      <input class="form__input form__input-big"
+             data-qa="keypress"
+             type="text"
+             value="Try &quot;Los Angeles&quot;">
```
TarasHoliuk, line 23: it should be a placeholder

src/style.css
```diff
@@ -1 +1,71 @@
 /* add styles here */
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  src: url(fonts/Avenir-Heavy.ttf);
+  src: url(fonts/Avenir-Book.ttf);
```
TarasHoliuk, line 6: 2 times the same URL

Also, you may write @font-face 2 times with different src and font-weight - to explicitly add the font with a higher font-weight (avenir-heavy)

TarasHoliuk (changes requested):
GJ :+1: 
Let's add a few small fixes

src/index.html
AnnSomko: done,thanks)

src/style.css
AnnSomko: change last url to Avenir.ttc;

src/index.html
```diff
@@ -11,6 +11,35 @@
...
+          action="/"
+    >
+      <img class="form__img-small"
+           src="images/Search.svg"
+           alt="search"
+      >
+
+      <input class="form__input form__input-small"
+             type="text"
+             placeholder="Try &quot;Los Angeles&quot;"
```
lerastarynets, line 41: attrCheklist

**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

lerastarynets (changes requested):
Almost there) Just several more things needs to be fixed

lerastarynets (approved):
You're making excellent progress!

PR #4078, opened 2024-02-19
added solution for two search bars

RostyslavKuzmych:


sarakhmen (approved):
I'm sure you can now find a lot of issues with your code yourself on the current stage of your learning :) so, if you have enough time please fix all the issues and move to the review again. But if no, you can skip it :)

PR #4098, opened 2024-02-24
Develop

Julia-na:
https://Julia-na.github.io/layout_search-bar-airbnb/
https://Julia-na.github.io/layout_search-bar-airbnb/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -39,7 +39,7 @@ input {
 
 .small {
   padding-left: 33px;
-  padding-bottom: 2.5px;
+  padding-bottom: 5px;
   height: 42px;
   box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
 }
--- a/src/style.css
+++ b/src/style.css
@@ -1,59 +1,55 @@
-/* add styles here */
-
 @font-face {
   font-family: Avenir, sans-serif;
-  src: url('./fonts/Avenir-Book.ttf');
+  src: url("./fonts/Avenir-Book.ttf");
   font-weight: 300;
 }
 
 @font-face {
   font-family: Avenir, sans-serif;
-  src: url('./fonts/Avenir-Book.ttf');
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,108 @@
...
+
+.small {
+  padding-left: 33px;
+  padding-bottom: 5px;
+  height: 42px;
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+}
+
+.big__bar,
+.small__bar {
```
TarasHoliuk, line 48: in such cases, it's better to add a common class to both elements i.e. `bar`, and use it to add all styles that you need for big bar for example.

Then for the small bar, you should add the class `bar bar__small` and in `bar__small` overwrite all properties you need (`background-size` and `background-position` in this case)

TarasHoliuk: Fix it in all similar cases

TarasHoliuk (changes requested):

src/style.css
```diff
@@ -1 +1,108 @@
...
+* {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Avenir, sans-serif;
+  font-size: 16px;
+}
+
+input {
```
TarasHoliuk, line 22: don't use tag selector, use `class` selector

TarasHoliuk (changes requested):

volodymyr-soltys97 (approved):
Great work 🔥

src/style.css
```diff
@@ -1 +1,98 @@
...
+}
+
+.bar__big::placeholder,
+.bar__small::placeholder {
+  font-family: Avenir, sans-serif;
+  display: inline-block;
+  color: #3d4e61;
+  font-style: normal;
+  font-weight: 300;
+  line-height: normal;
```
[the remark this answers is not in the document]
Julia-na: Thank you for the review!

PR #4100, opened 2024-02-26
Search bar try№1

VolodymyrLiubyn:
- [DEMO LINK](https://VolodymyrLiubyn.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://VolodymyrLiubyn.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,38 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <h1 class="visually-hidden">Search bar airbnb</h1>
+    <form action="url"
+    method="get"
+    name="big_search"
+    data-qa="big">
+      <div class="container">
+        <img src="./images/Search.svg" 
+        alt="searh_Logo"
+        class="search big_logo">
+        <input
+        data-qa="keypress"
+        type="searh"
+        id="big_form"
+        placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </div>
+    </form>
+    
+    <form action="url"
+      method="get"
+      name="small_search"
+      data-qa="small">
+      <div class="container">
+        <img src="./images/Search.svg" 
+        alt="searh_Logo"
+        class="search small_logo">
+        <input
+        type="searh"
+        id="small_form"
+        placeholder="Try &quot;Los Angeles&quot;"
+        >
+      </div>
+    </form>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,6 +11,38 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Search bar airbnb</h1>
+    <h1 class="visually-hidden">Search bar airbnb</h1>
+    <form action="url"
+    method="get"
+    name="big_search"
+    data-qa="big">
```
l4st1m0za, line 18:
```suggestion
    <form 
      action="url"
      method="get"
      name="big_search"
      data-qa="big"
    >
```
This is right way to style your code. Please consider it and change in **ALL** the same cases

src/style.css
```diff
@@ -0,0 +1,117 @@
+
+/* @font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-style: normal;
+  font-weight: 300;
+} */
```
l4st1m0za, line 7: Delete all the unnedeed comments.

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+  --biglogo-height: 19px;
+}
+
+.container {
+  position: relative;
+  margin-top: 20px;
+}
+
+#big_form,
+#small_form {
```
l4st1m0za, line 37: You don't need to use #. All you need is to call needed html elemet by its class.

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+  font-family: Avenir, sans-serif;
+  color: #3d4e61;
```
l4st1m0za, line 39: Move colors to separate variables

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+.visually-hidden {
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  margin: -1px !important;
+  border: 0 !important;
+  padding: 0 !important;
+  overflow: hidden !important;
+  clip: rect(0,0,0,0) !important;
+  white-space: nowrap !important;
```
l4st1m0za, line 89: You don't need to use !important

l4st1m0za (changes requested):
Good job, but you need to style your code in the right way, and also, try to pass the tasks with the things you have learned from the videos, if you have any questions you can always ask in FE chat

PR #4107, opened 2024-02-28
Develop

kgGugala:
- [DEMO LINK](https://kgGugala.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://kgGugala.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,96 @@
-/* add styles here */
+* {
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+:root {
+  --font-color: #3d4e61;
+}
+
+.page {
+  font-family: Avenir, sans-serif;
+}
+
+.form {
+  justify-content: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,96 @@
...
+  --font-color: #3d4e61;
+}
+
+.page {
+  font-family: Avenir, sans-serif;
+}
+
+.form {
+  justify-content: center;
+  margin: 20px auto;
```
yevhenii-pyl, line 26: Bad idea to use margins here.  Basically, you want to avoid `margin collapse` whenever you can to. 
Why don't you create a flex container for these forms and add `gap` to it?

src/style.css
```diff
@@ -1 +1,90 @@
...
+
+.form-container {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  width: 100%;
+  position: relative;
+  padding-left: 8px;
+  padding-right: 8px;
+  padding-top: 20px;
```
yevhenii-pyl, line 33: Having 2+ rules for padding requires simplification :) 
`padding: 20px 8px 0`

yevhenii-pyl (approved):
You go this! 🚀 

However, there is something off with demo: 
demo

I don't want to keep you for too long on basic tasks, so this one is approved in advance, but don't forget to double-check your demo link next time :)

PR #4121, opened 2024-03-04
Develop

Esther-Guimaraes:
- [DEMO LINK](https://Esther-Guimaraes.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Esther-Guimaraes.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,23 +10,37 @@
       http-equiv="X-UA-Compatible"
       content="ie=edge"
     />
-    <title>Document</title>
+    <title>Search bar airbnb</title>
     <link
       rel="stylesheet"
       href="style.css"
     />
   </head>
   <body>
-    <input
-      type="text"
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
-
-    <input
-      type="text"
-      data-qa="hover"
-      placeholder="Try “Los Angeles“"
-    />
+    <form
+      action="http://localhost:8080/"
+      method="post"
+      data-qa="big"
+    >
+      <input
+        class="input input__big"
+        type="search"
+        data-qa="keypress"
+        placeholder="Try “Los Angeles“"
+      />
+    </form>
+    
+    <form
+      action="http://localhost:8080/"
+      method="post"
+      data-qa="small"
+    >
+      <input
+        class="input input__small"
+        type="search"
+        data-qa="hover"
+        placeholder="Try “Los Angeles“"
+      />
+    </form>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -1 +1,86 @@
...
+  font-family: Avenir-Book;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: Avenir-Heavy;
+  src: url(./fonts/Avenir-Heavy.ttf);
+}
+
+html {
```
jv-aquino, line 16: you could also use :root but the difference isn't relevant really

jv-aquino (changes requested):
Esther, your code is nice, and your HTML is really organized, but there's some changes I'd like you to make:

1. Add CSS classes to the relevant elements and also use BEM with a feature to avoid much repetition (I think you can guess what I'm talking about)
2. there's some repeating code that you'll also fix with 1. so pay attention
3. for some reason the small input is strange on my machine, if it's just here please tell me 
on Mate Academy chat:
image

src/index.html
```diff
@@ -10,23 +10,39 @@
...
   </head>
   <body>
-    <input
-      type="text"
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
+    <form
+      class="search"
+      action="http://localhost:8080/"
```
EdPirro, line 22: Avoid setting `action="http://localhost:8080/"` it breaks in the deployed site

src/style.css
```diff
@@ -1 +1,70 @@
...
+  width: 100%;
+  box-sizing: border-box;
+  background-image: url(/src/images/Search.svg);
+  background-repeat: no-repeat;
+  color: var(--fontcolor);
+  border: 1px solid var(--colorgrey);
+  border-radius: 4px;
+  outline: none;
+  box-shadow: 0 1px 8px #3d4e611a;
+  font-family: Avenir, sans-serif;
```
EdPirro, line 36: It is considered best practices to apply the "default" font to the `body` tag so that every element that doesn't specify a font already uses it by default

src/style.css
```diff
@@ -1 +1,70 @@
...
+  background-position: 13px center;
+  background-size: 11px 11px;
+  font-size: 14px;
+}
+
+.search__input--small:hover,
+.search__input--big:hover {
+  box-shadow: 0 3px 8px 0 var(--dropcolor);
+  font-family: Avenir-Book, sans-serif;
+}
```
EdPirro, line 64: If both modifiers apply the same style you should consider moving them to the block/element instead

EdPirro (approved):
Great job, @Esther-Guimaraes! Very clean and straightforward code. I added a few suggestions for you to keep in mind for the next tasks (or implement in this you if you feel like it)

PR #4123, opened 2024-03-06
add searchbar

vinasxz:
- [DEMO LINK](https://vinasxz.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://vinasxz.github.io/layout_search-bar-airbnb/report/html_report/)
- [X] Icon implemented using background-image CSS property
- [X] All `Typical Mistakes` from `BEM` lesson theory are checked.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1123,9 +1123,9 @@
       "dev": true
     },
     "@mate-academy/scripts": {
-      "version": "1.7.2",
-      "resolved": "https://registry.npmjs.org/@mate-academy/scripts/-/scripts-1.7.2.tgz",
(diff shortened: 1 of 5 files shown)


jv-aquino (changes requested):
Overall your CSS code is really great as well as the structure, but there's a small tech problem that is not supposed to happen: 
The placeholder should leave only if the person typed something, if the input is empty it should still show "Try Los Angeles", like this:
image

pedro-ruas (approved):
I see you fixed previous requests and your code seems perfect. Great job!

A side note that could help you in the future, but not mandatory of course, you could resort to using SCSS instead of CSS and then use nesting for your classes and avoid repetition

PR #4129, opened 2024-03-06
add task solution

eduvinagre:
- [DEMO LINK](https://eduvinagre.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://eduvinagre.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,116 @@
-/* add styles here */
+:root {
+  --custom-color: rgba(61, 78, 97, 1);
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
+
+* {
+  box-sizing: border-box;
+  font-family: Avenir, sans-serif;
+  color: var(--custom-color);
+  font-size: 16px;
+}
+
+.search-form--large {
+  margin-bottom: 20px;
+  margin-top: 20px;
+}
+
+.search-form__label--large {
+  background-size: 19px 19px;
+  background-position: 26px center;
+}
+
+.search-form__input {
+  width: 100%;
+  height: 100%;
+
+  border: 1px solid rgba(225, 231, 237, 1);
+  border-radius: 4px;
+  outline: none;
+
+  font-weight: 300;
+}
+
+.search-form__input::placeholder {
+  color: var(--custom-color);
+}
+
+.search-form__input--small:not(
+    .search-form__input--small:focus,
+    .search-form__input--small:hover
+  )::placeholder {
+  text-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.1);
+}
+
+.search-form__input--large {
+  height: 70px;
+  padding-left: 62px;
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,116 @@
...
+}
+
+.search-form__input::placeholder {
+  color: var(--custom-color);
+}
+
+.search-form__input--small:not(
+    .search-form__input--small:focus,
+    .search-form__input--small:hover
+  )::placeholder {
```
jv-aquino, line 47: nice!

jv-aquino (changes requested):
Great job Eduardo!!! For real you progressed SO MUCH in BEM, CSS in general and code organization / cleaness. There's just a thing that's not related to code, but with the mockup (we can't approve it if it's not following, even if the code is Amazing like yours) that you need to fix:
image
There's some difference that I think is related to the font or font-weight when hovered / focused, the mockup above is different from it:
image

src/style.css
```diff
@@ -1 +1,116 @@
...
+:root {
+  --custom-color: rgba(61, 78, 97, 1);
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-style: normal;
+  font-weight: 300;
+}
```
EdPirro, line 10: You should use a `@font-face` for all the fonts you'll use and link the .ttf file you have to it [Reference.](https://github.com/mate-academy/layout_search-bar-airbnb/tree/master#:~:text=use%20%40font%2Dface%20for%20fonts) Just like you did here you must also do for Avenir-Heavy

src/style.css
EdPirro: Hey, @eduvinagre I think the point here is that your search bar is not using the Avenir-Heavy font; you need to link it to your .ttf file using `@font-face`. Here you can see since you don't define Avenir-Heavy the site defaults to sans-serif:

Your search-bar with Avenir-Heavy:
image

Without it:
image

And only with Avenir-Heavy:
image

Avenir-Heavy does not exist in html by default, that is why you need to use `@font-face` to import it

EdPirro (changes requested):
Great effort, @eduvinagre! You're almost there! Please go through my suggested changes and, as always, if you have any questions just let me know!

src/style.css
```diff
@@ -1 +1,116 @@
...
+
+* {
+  box-sizing: border-box;
+  font-family: Avenir, sans-serif;
+  color: var(--custom-color);
+  font-size: 16px;
+}
```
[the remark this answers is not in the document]
eduvinagre: Absolutely right! My bad! I had already changed that in the last fork, but I missed it this time for some reason.

eduvinagre:
Guys, I think that there was something about my Google Chrome and that's why I couldn't figure out what you guys were pointing out to me. The css was rendering with all the letters even though I hadn't been linking it on my code. Hope you guys can see it as I was seeing now.

pedro-ruas (approved):
I see you corrected everything the other mentors asked you to, and your code looks amazing, very well written!

PR #4179, opened 2024-03-22
add task solution

yuliaPel:
- [DEMO LINK](https://yuliaPel.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://yuliaPel.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,16 +17,46 @@
     />
   </head>
   <body>
-    <input
-      type="text"
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
+    <div class="search-block">
+      <div class="search-block__form">
+        <form
+          action="#"
+          data-qa="big"
+          class="search search--big"
+        >
+          <label
+            for="search-main"
+            class="search__icon search__icon--big"
+          ></label>
+          <input
+            class="search__input"
+            id="search-main"
+            type="text"
+            data-qa="keypress"
+            placeholder="Try “Los Angeles“"
+          />
+        </form>
+      </div>
 
-    <input
-      type="text"
-      data-qa="hover"
-      placeholder="Try “Los Angeles“"
-    />
+      <div class="search-block">
+        <form
+          action="#"
+          data-qa="small"
+          class="search-block__form search search--small"
+        >
+          <label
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -17,16 +17,46 @@
...
+      <div class="search-block__form">
+        <form
+          action="#"
+          data-qa="big"
+          class="search search--big"
+        >
+          <label
+            for="search-main"
+            class="search__icon search__icon--big"
+          ></label>
```
polosanya, line 30: Better wrap the input into label

src/style.css
```diff
@@ -1 +1,88 @@
...
+.search__icon--big {
+  top: 24.2px;
+  left: 25.2px;
+  height: 19px;
+  width: 19px;
+}
+
+.search__icon--small {
+  top: 14.2px;
+  left: 12.2px;
```
polosanya, line 71: Do you really need to use float values?

polosanya (changes requested):
Placeholder shouldn't change on hover. Also, let's make all search bar clickable, not just text
search

src/index.html
```diff
@@ -17,16 +17,48 @@
...
   </head>
   <body>
-    <input
-      type="text"
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
+    <div class="search-block">
+      <div class="search-block__form">
+        <form
```
polosanya, line 22: I think div is redundant as you already have a container for form

src/index.html
```diff
@@ -17,16 +17,48 @@
...
-    />
+      <div class="search-block__form">
+        <form
+          action="#"
+          data-qa="small"
+          class="search search--small"
+        >
+          <label
+            for="search-secondary"
+            class="search__label search__label--small"
```
polosanya, line 50: There is no need to create big and small modifiers. Just let one of them be the default one, so you need only one modifier

polosanya (approved):
Nice! I'm approving it, but please pay attention to the comments)

src/index.html
yuliaPel: Thank you! I corrected my mistakes)

PR #4193, opened 2024-03-27
added task solution

e1useteenoff:
- [DEMO LINK](https://e1useteenoff.github.io/layout_search-bar/)
- [TEST REPORT LINK](https://e1useteenoff.github.io/layout_search-bar/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,95 @@
 /* add styles here */
+* {
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+.form--big,
+.form--small {
+  position: relative;
+  margin-top: 20px;
+  transition: all 0.3s ease;
+}
+
+.input {
+  width: 100%;
+  box-sizing: border-box;
+  border: 1px solid #e1e7ed;
+}
+
+.input--small {
+  height: 42px;
+  padding-left: 33px;
+}
+
+.input--big {
+  height: 70px;
+  padding-left: 62px;
+}
+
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -1 +1,95 @@
...
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+.form--big,
+.form--small {
```
TarasHoliuk, line 13:
```suggestion
.form {
```
that's why you have common `form` class - to apply shared styles

src/style.css
```diff
@@ -1 +1,95 @@
...
+  position: relative;
+  margin-top: 20px;
+  transition: all 0.3s ease;
+}
+
+.input {
+  width: 100%;
+  box-sizing: border-box;
```
TarasHoliuk, line 21: it's better to set in `*` selector

src/style.css
TarasHoliuk: fix same everywhere

TarasHoliuk (changes requested):

src/style.css
```diff
@@ -1 +1,95 @@
+/* stylelint-disable comment-empty-line-before */
```
TarasHoliuk, line 1: seems redundant
why did you add it?

src/style.css
TarasHoliuk: it shouldn't break anything. At least, if everything looks the same. Anyway, let's keep it as it is, just take it into account in the future

TarasHoliuk (approved):
Approved with minor comment just FYI :heavy_check_mark:

PR #4196, opened 2024-03-31
add solution

yaroslav1177:
- [DEMO LINK](https://yaroslav1177.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://yaroslav1177.github.io/layout_search-bar-airbnb/report/html_report/)

❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,86 @@
 /* add styles here */
+body {
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 300;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 800;
+}
+
+:root {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,86 @@
...
+  color: #3d4e61;
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+  box-shadow: 0 1px 8px 0 #3d4e611a;
+}
+
+.big {
+  height: 70px;
+  background: var(--icon-big);
+  padding-left: 62px;
```
polosanya, line 41: Instead of having big and small input, you can make one of them the default one, and add modifier to another

polosanya (changes requested):
Good job! Let's improve your solution a bit)
Placeholder shouldn't change on hover
image

src/style.css
```diff
@@ -1 +1,81 @@
 /* add styles here */
```
roman-mirzoian, line 1: Do not leave unnecessary comments.

roman-mirzoian (approved):
Good job!

PR #4209, opened 2024-04-07
add solution

RuslanHnatyshyn:
- [DEMO LINK](https:///RuslanHnatyshyn.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https:///RuslanHnatyshyn.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,56 @@
 /* add styles here */
+
+@font-face {
+  font-family: Avenir;
+  src:
+    url('/src/fonts/Avenir-Book.ttf') format('truetype'),
+    url('/src/fonts/Avenir-Heavy.ttf') format('truetype');
+}
+
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  font-family: Avenir, sans-serif;
+}
+
+.search-bar {
+  margin: 20px 8px;
+}
+
+.search-bar__input {
+  box-sizing: border-box;
+  font-family: Avenir, sans-serif;
+  font-weight: 300;
+  cursor: pointer;
+  color: #3d4e61;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,56 @@
...
+  background-image: url(/src/images/Search.svg);
+  background-repeat: no-repeat;
+  background-position-x: 26px;
+  background-position-y: center;
+  height: 70px;
+  font-size: 16px;
+  padding-left: 62px;
+}
+
+.search-bar__input--small {
```
polosanya, line 47: There is no need to create two different modifiers. You can make one of search bars the default one, and add modifier only for another one

polosanya (changes requested):
Pay attention to the design.
The text should be bold on focus. Also, the whole input should have a different background
image
image

src/style.css
```diff
@@ -1 +1,66 @@
...
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+}
+
+.search-bar__input:hover {
+  box-shadow: 0 3px 8px #3d4e6133;
+}
+
+.search-bar__input--big::placeholder,
+.search-bar__input::placeholder {
+  font-family: Avenir, sans-serif;
```
polosanya, line 53: You can set it for the whole page and inherit where needed

polosanya (changes requested):
Almost done!
Icon shouldn't disappear on focus. Also, the placeholder shouldn't change
image

polosanya (approved):
Well done! I'm approving it, but please pay attention that the placeholder shouldn't change on focus
image

RuslanHnatyshyn:
Thank you for your comments.

PR #4231, opened 2024-04-17
Develop

lyarosh:
- [DEMO LINK](https://lyarosh.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://lyarosh.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,59 @@
-/* add styles here */
+.search-bar {
+  display: flex;
+  width: 100%;
+  border-spacing: 20px;
+}
+
+label {
+  background: url(images/Search.svg);
+}
+
+.search-bar[data-qa='big'] :default {
+  margin-top: 20px;
+  font-weight: 300;
+  height: 70;
+}
+
+.search-bar[data-qa='big'] :hover {
+  box-shadow: 0 4px 4px 0 #00000040;
+}
+
+.search-bar[data-qa='big'] :focus {
+  box-shadow: 0 4px 4px 0 #00000040;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
+  font-weight: bold;
+  text-shadow: 1px 4px 4px rgba(0, 0, 0, 0.25);
+}
+
+id-сity {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,85 @@
-/* add styles here */
+.search-bar {
+  display: flex;
+  margin-top: 20px;
+  margin-bottom: 20px;
```
anastasiia-tilikina, line 4: don't use both top and bottom margins to avoid margin collapse.
choose one and be consistent.
for example to add spacing between elements you can add for them only margin top

src/style.css
```diff
@@ -1 +1,85 @@
...
+  background: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  position: absolute;
+  height: 11px;
+  width: 11px;
+  margin-left: 13px;
+  top: 15px;
+  background-size: cover;
+}
```
anastasiia-tilikina, line 33: these two selectors have a lot in common you can create one class common for this element add styles for big one and then override for small those you need
for instance:

```css
.label::before {
  content: '';
  background: url(images/Search.svg);
  background-repeat: no-repeat;
  background-position: center;
  position: absolute;
  height: 19px;
  width: 19px;
  margin-left: 26px;
  top: 25px;
  background-size: cover;
}

.label-small:before {
  height: 11px;
  width: 11px;
  margin-left: 13px;
  top: 15px;
}
```

src/style.css
```diff
@@ -1 +1,85 @@
...
+
+.search-bar--big {
+  height: 70px;
```
anastasiia-tilikina, line 36: don't hardcode height use proper paddings instead

anastasiia-tilikina (changes requested):
GJ!

left some suggestions for you

src/style.css
vadiimvooo: Agreed with Anastasia

vadiimvooo (changes requested):
Almost done, just fix all comments

PR #4259, opened 2024-04-27
Add search bar airbnb

siefimov:
- [DEMO LINK](https://siefimov.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://siefimov.github.io/layout_search-bar-airbnb/report/html_report/)

- [ x ] Icon implemented using background-image CSS property
- [ x ] Inputs are written inside of 'form' tag with correctly passed attributes
- [ x ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ x ] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -14,25 +14,32 @@
 }
 
 body {
-  display: flex;
-  flex-direction: column;
+  /* display: flex;
+  flex-direction: column; */
   margin: 0;
   font-weight: 300;
 }
 
-input {
+/* input {
   border: none;
   outline: none;
+} */
+
+.form {
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
 }
 
-input::placeholder {
+.form__search-bar--large::placeholder,
+.form__search-bar--small::placeholder {
   font-family: Avenir, sans-serif;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,92 @@
...
+  src: url(./fonts/Avenir-Heavy.ttf);
+}
+
+@font-face {
+  font-family: 'Avenir TTC';
+  src: url(./fonts/Avenir.ttc);
+}
+
+body {
+  margin: 0;
```
volodymyr-soltys97, line 17: No need to reset margin for the body in this task

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You don't need to hide the search icon when input is focused
image
image

src/style.css
siefimov: I've fix icon's url, so now the search icon have to be displayed when input is focused.
Margin for the body is removed.

volodymyr-soltys97 (approved):
Well done! 🔥

PR #4264, opened 2024-04-29
add task solution

AndriiZakharenko:

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,23 +10,34 @@
       http-equiv="X-UA-Compatible"
       content="ie=edge"
     />
-    <title>Document</title>
+    <title>Search bar airbnb</title>
     <link
       rel="stylesheet"
       href="style.css"
     />
   </head>
-  <body>
-    <input
-      type="text"
-      data-qa="keypress"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -10,23 +10,34 @@
...
-      type="text"
-      data-qa="hover"
-      placeholder="Try “Los Angeles“"
-    />
+    <form
+      class="form"
+      data-qa="small"
+    >
+      <input
+        class="form__input input--small"
```
polosanya, line 37: There is actually no need to add two separate modifiers (big and small).
You can just make one of them be the default one

src/style.css
```diff
@@ -1 +1,94 @@
...
+}
+
+.form__input::placeholder {
+  color: #3d4e61;
+}
+
+.input--big {
+  height: 70px;
+  font-size: 16px;
+  line-height: 21.86px;
```
polosanya, line 48: If you need to use float values - you are probably doing something wrong

polosanya (changes requested):
Almost done! Don't forget to add links to the PR description.
Placeholder shouldn't change on focus
image

PR #4282, opened 2024-05-08
add search-bar solution

FaiHamid:
- [DEMO LINK](https://FaiHamid.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://FaiHamid.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,63 @@
 /* add styles here */
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf), url(./fonts/Avenir-Heavy.ttf);
+}
+
+html {
+  font-family: Avenir, sans-serif;
+  font-size: 16px;
+  color: #3d4e61;
+  font-weight: 300px;
+}
+
+body {
+  margin: 8px;
+}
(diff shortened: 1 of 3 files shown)


dinaminahi (changes requested):
Good job 👍
Let's do some fixes
The focus state should be without border and search icon should be visible. Please, check the mockups
image

dinaminahi (changes requested):
Better 👍 
But still on focus it is different from mockup
the text is bolder and have shadow and the field should be like in hover state
image

src/style.css
```diff
@@ -1 +1,123 @@
...
+
+.search-form__input--second {
+  height: 42px;
+  background-size: 11px;
+  background-position: left 12px top 14px;
+  padding: 0 0 0 33px;
+
+  font-family: inherit;
+  font-size: 14px;
+  font-weight: 400px;
```
roman-mirzoian, line 48: Its a default value it is not necessary to specify it.
```suggestion
```

roman-mirzoian (approved):
Well done!

PR #4298, opened 2024-05-18
add task solution

EqVert:
- [DEMO LINK](https://EqVert.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://EqVert.github.io/layout_search-bar-airbnb/report/html_report/)

- [X] Icon implemented using background-image CSS property
- [X] Inputs are written inside of 'form' tag with correctly passed attributes
- [X] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [X] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,60 @@
-/* add styles here */
+* {
+  padding: 0;
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url('./fonts/Avenir.ttc') format('truetype');
+  font-weight: 500;
+}
+
+input {
+  display: flex;
+  width: 100%;
+  color: #3d4e61;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -1 +1,60 @@
...
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url('./fonts/Avenir.ttc') format('truetype');
+  font-weight: 500;
+}
+
+input {
```
volodymyr-soltys97, line 12: Use only class selector for inputs, fix it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to remove default border when input is focused 
image

PR #4334, opened 2024-06-17
Develop

Artem-Brui:
- [DEMO LINK](https://Artem-Brui.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Artem-Brui.github.io/layout_search-bar-airbnb/report/html_report/)

Hello, 
I had an issue with attributes. 
There are conditions like:
- add attribute data-qa="big" for big search form, and data-qa="small" for small
- add attribute data-qa="keypress" to input in big search form
but....

when I'm creating an attribute like "big keypress" - test is not passing. As I understood, because test looking for exactly an item with data-qa="big", but not data-qa="big keypress". So I didn't use data-qa="keypress" here. I know than I can target to item with element by using next: .myClass[data-qa~='keypress'] and I used it in my scss. 

If I made a mistake, please let me know about that. Thanks in advance.


lerastarynets (approved):
You're doing a fantastic job, keep going!

PR #4339, opened 2024-06-18
Search Bar

Kateryna99:
- [DEMO LINK](https://Kateryna99.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Kateryna99.github.io/layout_search-bar-airbnb/report/html_report/)

- [x ] Icon implemented using background-image CSS property
- [x ] Inputs are written inside of 'form' tag with correctly passed attributes
- [ x] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [x ] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,16 +17,35 @@
     />
   </head>
   <body>
-    <input
-      type="text"
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
-
-    <input
-      type="text"
-      data-qa="hover"
-      placeholder="Try “Los Angeles“"
-    />
+    <div class="app">
+      <div class="app__wrapper">
+        <form
+          class="search-bar"
+          data-qa="big"
+        >
+          <label class="search-input">
+            <input
+              type="text"
+              class="search-input__item"
+              placeholder="Try “Los Angeles”"
+              data-qa="keypress"
+            />
+          </label>
+        </form>
+        <form
+          class="search-bar"
+          data-qa="small"
+        >
+          <label class="search-input search-input--small">
+            <input
+              type="text"
+              class="search-input__item search-input__item--small"
+              placeholder="Try “Los Angeles”"
+              data-qa="hover"
+            />
+          </label>
+        </form>
+      </div>
+    </div>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,127 @@
-/* add styles here */
+@font-face {
+  font-family: Avenir;
+  src: url('fonts/Avenir.ttc') format('truetype');
+  font-weight: 300;
+  font-style: normal;
+  font-display: swap;
+}
+
+@font-face {
+  font-family: Avenir;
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -1 +1,127 @@
...
+:root {
+  --color-white: #fff;
+}
+
+.app {
+  font-weight: 300;
+}
+
+.app__wrapper {
+  padding: 20px 8.5px 20px 8px;
```
ArturErmolenko, line 42:
```suggestion
  padding: 20px 8px 20px 8px;
```
It's better to use integers without a dot

src/index.html
```diff
@@ -17,16 +17,35 @@
...
-      placeholder="Try “Los Angeles“"
-    />
-
-    <input
-      type="text"
-      data-qa="hover"
-      placeholder="Try “Los Angeles“"
-    />
+    <div class="app">
+      <div class="app__wrapper">
```
ArturErmolenko, line 21: I believe we don't need these additional wrappers, so we increase the number of elements in the DOM tree, which is not good for performance.

Could we move (if needed) styles to the body?
Also, the class app is not the best choice , better use body or form etc

src/index.html
```diff
@@ -17,16 +17,35 @@
...
+        <form
+          class="search-bar"
+          data-qa="big"
+        >
+          <label class="search-input">
```
ArturErmolenko, line 26:
```suggestion
          <label class="search-label">
```
it's suits better

src/index.html
```diff
@@ -17,16 +17,35 @@
...
+            <input
+              type="text"
+              class="search-input__item"
```
ArturErmolenko, line 29: and this is input ;)
```suggestion
              class="search-input"
```

ArturErmolenko (changes requested):
Great start 🔥 
Let's fix small comments about BEM , on your Demo link there is a problem with small search( placeholder & icon are cropped) could you double check it plz 🐝 
Screenshot 2024-06-19 at 12 26 13

src/style.css
Kateryna99: Yes ! I 100% agree with you but I could not pass tests otherwise

src/index.html
Kateryna99: In this case I do not use label separately from input, input is a part of label , like a solid BEM block label>input

src/index.html
Kateryna99: check answer above

src/style.css
```diff
@@ -1 +1,125 @@
...
+  font-weight: 900;
+  font-style: normal;
+  font-display: swap;
+}
+
+/* FONTS */
+
+* {
+  margin: 0;
+  padding: 0;
```
BudnikOleksii, line 29: Don't set margins/paddings, etc to all elements, reset only for the tags that you actually need. For example, div/span doesn't have any margins or padding, so you don't need to reset them

BudnikOleksii (approved):
👍

PR #4342, opened 2024-06-19
add task solution first attempt

Uandrik:
- [DEMO LINK](https://Uandrik.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Uandrik.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,69 @@
-/* add styles here */
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Book.ttf);
+  font-weight: 400;
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: bold;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.page__body {
+  margin-top: 0;
+}
+
+.search-bar--big {
+  margin-top: 20px;
+  margin-bottom: 20px;
+}
+
+.search-bar__input {
+  width: 100%;
+  height: 100%;
+  border: 1px solid rgba(225, 231, 237, 1);
+  border-radius: 4px;
+
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,69 @@
...
+.page__body {
+  margin-top: 0;
+}
+
+.search-bar--big {
+  margin-top: 20px;
+  margin-bottom: 20px;
+}
+
+.search-bar__input {
```
TarasHoliuk, line 26: according to the mockup, there should be a gradient background on the focus. You can add 2: image and gradient as a background - google how to do it properly. Or add image using tag

TarasHoliuk (changes requested):
GJ 👍 
One improvement is required

src/style.css
```diff
@@ -1 +1,70 @@
...
+}
+
+.search-bar__input:hover {
+  box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.25);
+}
+
+.search-bar__input:focus {
+  outline: none;
+  font-weight: 900;
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%);
```
TarasHoliuk, line 46: this property overwrites background-image property, so you should add it here as well:

image

Here are some related links:
https://www.geeksforgeeks.org/css-combine-background-image-with-gradient-overlay/

https://stackoverflow.com/questions/2504071/how-do-i-combine-a-background-image-and-css3-gradient-on-the-same-element

TarasHoliuk (changes requested):

src/style.css
```diff
@@ -1 +1,81 @@
...
+  background-image: url(images/Search.svg);
```
TarasHoliuk, line 47: Unfortunately, it doesn't work like that

You should write both: the gradient and image in one property, separated by coma (exactly as in my example)
image

Also, check examples in links I've sent

TarasHoliuk (changes requested):
Almost, but now gradient doesn't work at all 😢 
Small fix is required

PR #4352, opened 2024-06-28
Develop

Miranna472:


src/index.html
```diff
@@ -17,16 +17,27 @@
...
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
-
-    <input
-      type="text"
-      data-qa="hover"
-      placeholder="Try “Los Angeles“"
-    />
+    <div class="search-bar-container">
```
volodymyr-soltys97, line 20: Redundant
```suggestion
```

src/style.css
```diff
@@ -1 +1,121 @@
...
+  font-weight: 700;
+  font-style: normal;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Avenir, Arial, Helvetica, sans-serif;
```
volodymyr-soltys97, line 27: Move font styles to the html selector

src/style.css
```diff
@@ -1 +1,121 @@
...
+  box-sizing: border-box;
+  font-size: 16px;
+  text-align: left;
+  font-weight: 300;
+  font-style: normal;
+  margin-top: 0;
```
volodymyr-soltys97, line 33: Redundant
```suggestion
```

src/style.css
```diff
@@ -1 +1,121 @@
...
+
+.search-bar.big {
+  width: 100%;
+  display: flex;
+  align-items: center;
+  margin-bottom: 20px;
+  margin-top: 20px;
+}
+
+.search-bar.big input {
```
volodymyr-soltys97, line 51: Use only class selector instead of input, fix it everywhere

src/style.css
```diff
@@ -1 +1,121 @@
...
+.search-bar:hover {
+  border-color: #3d4e61;
+  font-size: 16px;
+  font-weight: 300;
+}
+
+.search-bar.big input:focus {
+  font-size: 16px;
+  font-weight: 900;
+  line-height: 21.86px;
```
volodymyr-soltys97, line 108:
```suggestion
  line-height: 22px;
```

src/style.css
```diff
@@ -1 +1,121 @@
...
+  text-align: left;
```
volodymyr-soltys97, line 109: Default value, you can just remove it everywhere
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
You need to fix height for the input, height must be 70px 
image

src/style.css
Miranna472: when I move everything to html, all the fonts fly off. That's why I left it as it was

Miranna472:
- [х] Icon implemented using background-image CSS property
- [х] Inputs are written inside of 'form' tag with correctly passed attributes
- [х] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [х] Code follows all the [Code Style Rules ❗️](./checklist.md)

src/style.css
```diff
@@ -1 +1,68 @@
...
+  font-style: normal;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Avenir, Arial, Helvetica, sans-serif;
+  box-sizing: border-box;
```
SanyaBratashchuk, line 28: it is common practice to apply same box-sizing for all elements

SanyaBratashchuk (approved):

PR #4355, opened 2024-07-01
Add solution

k-marchuk:
- [DEMO LINK](https://k-marchuk.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://k-marchuk.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,77 @@
-/* add styles here */
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+html * {
+  --main-text-color: #3d4e61;
+  --box-shadow-color: #3d4e6133;
+  --border-color: #e1e7ed;
+  --text-shadow-color: #00000040;
+
+  font-family: Avenir, sans-serif;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,77 @@
-/* add styles here */
+@font-face {
+  font-family: Avenir;
+  src: url(./fonts/Avenir-Book.ttf);
+}
+
+html * {
```
BudnikOleksii, line 6: Don't use * selector (it impacts performance)
Variables you can specify for the `:root` selector https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties. And font to `html` element

BudnikOleksii (approved):
Good work 👍 
Approved, but check the comment

src/style.css
k-marchuk: Thanks for the review, I got it! 
I've tried to add font-styles only to `html` element but _user agent stylesheet_ overrides it every time. Otherwise, I needed to write a  font-style for each `form` element separately.

PR #4359, opened 2024-07-02
Task Solution: Search bar

Matheusgroff:
- [DEMO LINK](https://Matheusgroff.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Matheusgroff.github.io/layout_search-bar-airbnb/report/html_report/)

I used the properties from figma but I don't know how to make the upper bar bigger without defining height and width (it asks to use 100% of the latter but if I do it they get the same sizes).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -30,7 +30,7 @@ ___
 
 ❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.
 
-- [ ] Icon implemented using background-image CSS property
-- [ ] Inputs are written inside of 'form' tag with correctly passed attributes
-- [ ] All `Typical Mistakes` from `BEM` lesson theory are checked.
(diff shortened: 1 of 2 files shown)


alysonvilela (changes requested):
- use `@font-face` for fonts

Please add the requirements correctly.
- On your tests seems like there is a padding-left and padding right missing on the input container.
- The space between the icon and the placeholder is incorrect
- The second icon should be smaller than the first one.
image

alysonvilela (changes requested):
Remember to always use the TEST REPORT to support you and keep the code exactly like is expected to be.

Here is things I got just from hit on https://matheusgroff.github.io/layout_search-bar-airbnb/report/html_report/

- Font is not the same as the reference
- Missing shadow boxes
- Missing container padding/input margin

alysonvilela (approved):

PR #4383, opened 2024-07-18
DEVELOP

Yatsomo:


Yatsomo:
The checks pass in VSC but fail on GitHub? How so?

src/style.css
```diff
@@ -1 +1,61 @@
...
+  background-image: url('./images/Search.svg');
+  background-repeat: no-repeat;
+  background-color: #fff;
+  border-radius: 4px;
+  box-shadow: 0 1px 8px rgba(61, 78, 97, 0.1);
+  outline: none;
+  box-sizing: border-box;
+  border: 1px solid #e1e7ed;
+  font-family: Avenir, sans-serif;
+  color: #3d4e61;
```
natalia-klonowska, line 30: you need to use ::placeholder pseudo-element to change color

src/style.css
```diff
@@ -1 +1,61 @@
...
+}
+
+.searchbar__big {
```
natalia-klonowska, line 33: try to use modifier instead

src/style.css
```diff
@@ -1 +1,61 @@
...
+  background-position: 25px 24px;
+  background-size: 19px 19px;
+  padding-left: 62px;
+  font-size: 16px;
+  line-height: 22px;
+}
+
+.searchbar__small {
+  height: 42px;
+  background-position: 12px 14px;
```
natalia-klonowska, line 44: it should be 13px 15px

src/style.css
```diff
@@ -1 +1,61 @@
...
+  padding-left: 33px;
+  font-size: 14px;
+  line-height: 19px;
+}
+
+.searchbar__input:hover {
+  box-shadow: 0 3px 8px rgba(61, 78, 97, 0.2);
+}
+
+.search__bar:focus {
```
natalia-klonowska, line 55: wrong class name

natalia-klonowska (changes requested):
typed text should have different font-weight 
image

src/style.css
Yatsomo: When its like that the test comes back negative. Should I still change it?

src/index.html
```diff
@@ -17,16 +17,30 @@
...
-      type="text"
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
+    <form
+      class="searchbar"
+      data-qa="big"
+    >
+      <input
+        class="searchbar__input searchbar__big"
```
Zibi95, line 25: Its great case to use modifier class name like this `searchbar__input searchbar__input--big`

src/index.html
```diff
@@ -17,16 +17,30 @@
...
+        type="text"
+        name="searchbar__bar searchbar__big"
```
Zibi95, line 27: This is weird value for name property. It is primarily used in form elements to identify input fields when they are submitted to a server.

src/style.css
```diff
@@ -1 +1,64 @@
...
+}
+
+@font-face {
+  font-family: Avenir;
+  src: url(fonts/Avenir-Heavy.ttf);
+  font-weight: 900;
+}
+
+.searchbar {
+  margin: 20px 0;
```
Zibi95, line 16: In this task its okay. But remember that if you want to configure the external geometry or positioning of a block, use separate classes or mix-ins to separate external properties (positioning, margins) from internal properties (padding). This approach helps avoid conflicts between element styles.

src/style.css
Zibi95: If it fails test then don't change it.

Zibi95 (approved):
Good job.

PR #4395, opened 2024-07-27
upload tarefa search-bar

14lucas-mendes:
[DEMO LINK](https://14lucas-mendes.github.io/layout_search-bar-airbnb/)
[TEST REPORT LINK](https://14lucas-mendes.github.io/layout_search-bar-airbnb/report/html_report/)


src/style.css
```diff
@@ -1 +1,62 @@
...
+  font-family: Avenir;
+  src: url(/src/fonts/Avenir.ttc);
+}
+
+input {
+  font-family: Avenir, sans-serif;
+  width: 100%;
+}
+
+.barra-big {
```
pedro-ruas, line 16: It is better to use the name in full English "barra" is a Portuguese word.
Also, "small" is a modifier, and therefore should be separated by 2 hyphens instead of 1. Like this: `.bar--big`
Same advice is valid for the other class "barra-small"

pedro-ruas (changes requested):
Small BEM changes needed

gitlherme (changes requested):
Hello, @14lucas-mendes! Really great job, but one of your tests are failing. Please check your small bar. As you can see below, its a little different than the reference. The color is not the same and I think that is making you test failing. Please check it and make sure that all your tests are passing sucessfully.


image

pedro-ruas (changes requested):
Lucas, you code is very well done, but you are failing one required test.
You can check the result at your TEST LINK, seems that the color for bar--small is wrong

pedro-ruas (approved):
Great job!

PR #4401, opened 2024-07-29
Develop

nex1994:
- [DEMO LINK](https://nex1994.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://nex1994.github.io/layout_search-bar-airbnb/report/html_report/)
-  [x] Icon implemented using background-image CSS property

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,23 +10,59 @@
       http-equiv="X-UA-Compatible"
       content="ie=edge"
     />
-    <title>Document</title>
+    <title>Mate academy | airbnb search bar</title>
     <link
       rel="stylesheet"
       href="style.css"
     />
+    <link
+      rel="stylesheet"
+      href="big-search-bar.css"
+    />
+    <link
+      rel="stylesheet"
+      href="small-search-bar.css"
+    />
   </head>
-  <body>
-    <input
-      type="text"
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
+  <body class="page">
+    <form
+      class="form_big page__form_big"
+      action="GET"
+      data-qa="big"
+    >
+      <label
+        for="big-bar"
+        class="label form_big__label"
+        data-qa="big-bar"
+      >
+        <input
+          class="label__input"
+          name="big-bar"
+          id="big-bar"
+          type="text"
+          data-qa="keypress"
+          placeholder="Try “Los Angeles“"
+        />
+      </label>
+    </form>
 
-    <input
-      type="text"
-      data-qa="hover"
-      placeholder="Try “Los Angeles“"
-    />
+    <form
+      class="form_small page__form_small"
+      action="GET"
+      data-qa="small"
+    >
+      <label
+        class="label form_small__label"
+        for="small-bar"
+      >
+        <input
+          class="label__input"
+          name="small-bar"
+          id="small-bar"
+          type="text"
+          placeholder="Try “Los Angeles“"
+        />
+      </label>
+    </form>
   </body>
 </html>
--- a/src/small-search-bar.css
+++ b/src/small-search-bar.css
@@ -1,0 +1,62 @@
+@font-face {
+  font-family: 'Avenir book';
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: 'Avenir heavy';
+  src: url(fonts/Avenir-Heavy.ttf);
+}
+
+.page__form_small {
+  margin-top: 20px;
+}
+
+.form_small {
+  display: flex;
+  align-items: center;
+  height: 40px;
+  background-image: url(images/Search.svg);
+  background-repeat: no-repeat;
+  background-size: 11px;
+  background-position: 12px 14px;
+  border: 1px solid rgba(225, 231, 237, 1);
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  border-radius: 4px;
+}
+
+.form_small__label {
+  display: flex;
+  align-items: center;
+  height: 100%;
+  width: 100%;
+}
+
+.form_small:hover {
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+}
+
+.form_small:focus {
+  box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.25);
+}
+
+.form_small .label__input {
+  position: relative;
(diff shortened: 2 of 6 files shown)


src/index.html
```diff
@@ -10,23 +10,59 @@
...
     />
+  </head>
+  <body class="page">
+    <form
+      class="form_big page__form_big"
+      action="GET"
+      data-qa="big"
+    >
+      <label
+        for="big-bar"
```
Zibi95, line 34: if you keep the input inside label tag you don't need id to connect them.

src/small-search-bar.css
```diff
@@ -0,0 +1,62 @@
+@font-face {
+  font-family: 'Avenir book';
+  src: url(fonts/Avenir-Book.ttf);
+}
+
+@font-face {
+  font-family: 'Avenir heavy';
+  src: url(fonts/Avenir-Heavy.ttf);
+}
```
Zibi95, line 9: You don't have to declare it twice.

src/small-search-bar.css
```diff
@@ -0,0 +1,62 @@
...
+
+.page__form_small {
+  margin-top: 20px;
+}
+
+.form_small {
+  display: flex;
+  align-items: center;
+  height: 40px;
```
Zibi95, line 18: I think that the element should have height for example input or label. Don't fix container size (if there is no such a requirement) rule is broken here.

src/small-search-bar.css
```diff
@@ -0,0 +1,62 @@
...
+
+.form_small .label__input::placeholder {
+  font-weight: 300;
+  color: rgba(61, 78, 97, 1);
+  text-decoration: none;
+}
+
+.form_small .label__input:focus {
+  border: none;
+}
```
Zibi95, line 62: Look at all this code its almost the same as for big-search-bar. Make use of modifier class and just create one block instead of 2.

src/small-search-bar.css
```diff
@@ -0,0 +1,62 @@
...
+.form_small .label__input {
+  position: relative;
+  left: 33px;
+  border: none;
+  font-family: 'Avenir book', Arial, Helvetica, sans-serif;
+  font-size: 14px;
+  font-weight: 300;
+  color: rgba(61, 78, 97, 1);
+  padding: 0;
+}
```
Zibi95, line 52: Don't increase the specificity of a selector unless necessary.

src/index.html
```diff
@@ -10,23 +10,59 @@
...
+        class="label form_big__label"
```
Zibi95, line 35: you don't use class name of `label` anywhere in stylesheets. Remove it and fix elements inside it

src/small-search-bar.css
Zibi95, line 1: Name the files according to the block name.

Zibi95 (changes requested):
Hello. I have left some comments in the code for this task. Please fix them and minimize any repeated code.

src/search-bar.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.form--big:focus,
+.form--small:focus {
+  box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.25);
+}
+
+.form--big:hover,
+.form--small:hover {
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
```
Zibi95, line 74: If it is for both cases then you could just use `.form` selector

src/index.html
```diff
@@ -10,23 +10,49 @@
...
+      class="form page__form form--big"
+      action="GET"
+      data-qa="big"
+    >
+      <label
+        class="label label--big"
+        data-qa="big-bar"
+      >
+        <input
+          class="label__input input--big"
```
Zibi95, line 34: You have used modifier (input--big) on input without using block (input) className. Its one of BEM mistakes.

src/index.html
```diff
@@ -10,23 +10,49 @@
...
     />
+  </head>
+  <body class="page">
+    <form
+      class="form page__form form--big"
+      action="GET"
+      data-qa="big"
+    >
+      <label
+        class="label label--big"
```
Zibi95, line 30: You have created so many blocks like label, form, page. This task can be done with using only one block.

src/search-bar.css
```diff
@@ -0,0 +1,83 @@
...
+  border: 1px solid rgba(225, 231, 237, 1);
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+}
+
+.form--small {
+  height: 40px;
+  background-size: 11px;
+  background-position: 12px 14px;
+  border: 1px solid rgba(225, 231, 237, 1);
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
```
Zibi95, line 31: These ones are also duplicated in form--big

src/search-bar.css
```diff
@@ -0,0 +1,83 @@
...
+
+.label--big {
+  height: 100%;
+  width: 100%;
+}
+
+.label--small {
+  height: 100%;
+  width: 100%;
+}
```
Zibi95, line 47: Those modifiers doesn't make difference

Zibi95 (changes requested):
Hey. There are still some BEM mistakes which are not allowed. Look into file with typical BEM mistakes and make sure you follow the tips from there 👍

src/index.html
```diff
@@ -10,23 +10,49 @@
...
-      placeholder="Try “Los Angeles“"
-    />
+    <form
+      class="form page__form form_small"
+      action="GET"
+      data-qa="small"
+    >
+      <label class="form__item1 form__item1_small">
+        <input
+          class="form__item2 form__item2_small"
```
Zibi95, line 50: those class names are bad. Give them more meaning like `form__label`, `form__input`

src/index.html
```diff
@@ -10,23 +10,49 @@
...
+      class="form page__form form_big"
+      action="GET"
+      data-qa="big"
+    >
+      <label
+        class="form__item1"
+        data-qa="big-bar"
+      >
+        <input
+          class="form__item2 form__item2_big"
```
Zibi95, line 34: There are many ways to name classes in BEM but for me the best one is when we use double hyphens for modifiers. `form__item2--big`

Zibi95 (approved):
Good job this time its okay. Only then class names are bad. I approve this but remember to give more meaningful class names.

PR #4413, opened 2024-08-06
Task solution

MateuszCieplak:
- [DEMO LINK] (https://MateuszCieplak.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK] (https://MateuszCieplak.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,16 +17,34 @@
     />
   </head>
   <body>
-    <input
-      type="text"
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
+    <form
+      class="search-form search-form__big"
+      data-qa="big"
+    >
+      <label class="search-form__label search-form__label__big">
+        <!-- prettier-ignore -->
+        <input
+          class="search-form__input search-form__input__big"
+          type="text"
+          data-qa="keypress"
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -17,16 +17,34 @@
...
-      type="text"
-      data-qa="keypress"
-      placeholder="Try “Los Angeles“"
-    />
+    <form
+      class="search-form search-form__big"
+      data-qa="big"
+    >
+      <label class="search-form__label search-form__label__big">
+        <!-- prettier-ignore -->
```
natalia-klonowska, line 25: why ignore prettier?

natalia-klonowska (changes requested):
you have border radius only on focus and on focus search bars shouldn't have outline and bigger one need text-shadow

natalia-klonowska (changes requested):
text-shadow should look different 
image

src/index.html
MateuszCieplak: It is because when I made "npm run lint", the placeholder change from "Try &quot;Los Angeles&quot;" into 'Try "Los Angeles" ' and i could not came through it. So I found a solution on internet and it works

natalia-klonowska (approved):

PR #4419, opened 2024-08-09
add index.html and style.css files

YevhenRDoIT:
Interesting issue with placeholder value placeholder='Try "Los Angeles"'. So when I run command 'npm run lint' it doesn't pass this way because it asks for double qoutes around "Try "Los Angeles"" but this way it still doesn't pass. So i decided to do this way: placeholder="Try Los Angeles" and command 'npm run lint' passed it. But the command 'npm test' says that there are three fails because of no double qoutes around text "Los Angeles". Yes I also have tried this way: placeholder=`Try ${"Los Angeles"}` but it doesn't work here.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -25,8 +25,8 @@ ___
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_search-bar-airbnb/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_search-bar-airbnb/report/html_report/)
+- [DEMO LINK](https://YevhenRDoIT.github.io/layout_search-bar-airbnb/)
+- [TEST REPORT LINK](https://YevhenRDoIT.github.io/layout_search-bar-airbnb/report/html_report/)
 
 ❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.
 
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,121 @@
 /* add styles here */
+* {
+  padding: 0;
+  box-sizing: border-box;
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 300;
+  font-style: normal;
+  src: url('./fonts/Avenir-Book.ttf') format('truetype');
+}
+
+@font-face {
+  font-family: Avenir;
+  font-weight: 600;
+  font-style: normal;
+  src: url('./fonts/Avenir-Heavy.ttf') format('truetype');
+}
+
+body {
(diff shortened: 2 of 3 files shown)


etojeDenys (changes requested):
pls, feel free to ask for some help with those quotes in the chat

YevhenRDoIT:
I found this syntax way: placeholder="Try &quot;Los Angeles&quot;" but still I can't pass the test.

YevhenRDoIT:
I found syntax ways to solve it, as follows: placeholder="Try &quot;Los Angeles&quot;" and when I run this command -> 'npm run lint' the following syntax automatically changes to placeholder='Try "Los Angeles"' and in result test can't be passed. Sorry but its really wierd and I think that something is wrong while lint checks for errors following syntax code: placeholder="Try &quot;Los Angeles&quot;". So I had to use following syntax to at least push it: placeholder="Try 'Los Angeles'"

YevhenRDoIT:
I just updated value of attribute 'placeholder' in index.html. And it shows that tests are passed. Please check it and approve it. Thank you )

YevhenRDoIT (commented):
updates in placeholder

readme.md
```diff
@@ -25,8 +25,8 @@ ___
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_search-bar-airbnb/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_search-bar-airbnb/report/html_report/)
+- [DEMO LINK](https://YevhenRDoIT.github.io/layout_search-bar-airbnb/)
+- [TEST REPORT LINK](https://YevhenRDoIT.github.io/layout_search-bar-airbnb/report/html_report/)
```
volodymyr-soltys97, line 29: Add these links to the PR description 
image

src/style.css
```diff
@@ -1 +1,121 @@
...
+@font-face {
+  font-family: Avenir;
+  font-weight: 600;
+  font-style: normal;
+  src: url('./fonts/Avenir-Heavy.ttf') format('truetype');
+}
+
+body {
+  font-family: Avenir, sans-serif;
+  font-size: 16px;
```
volodymyr-soltys97, line 23: Move font styles to the html selector

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/style.css
YevhenRDoIT: ok

volodymyr-soltys97 (approved):
Hi, this comment still not fixed, I approved your solution, but in the future you should to add these links to the PR description  
image

PR #4423, opened 2024-08-10
add solution

ghost:
### Requirements:
- use images from [src/images](src/images)
- there must be two search bars
- search bar must stretch the full width
- distance between two search bars must be 20px
- the big search bar must have top indent 20px
- follow styles from the mock
- default `font-weight` must be 300
- a search bar has 3 state default, `hover` and `focus`
- don't use JavaScript
- use `@font-face` for fonts


vadiimvooo (approved):
Well done

PR #4435, opened 2024-08-17
add task solution

VolodymyrV7:


vadiimvooo (changes requested):
Please try to pass all tests. Feel free to ask for help in fe_chat.

VolodymyrV7:
х Icon implemented using background-image CSS property
х Inputs are written inside of 'form' tag with correctly passed attributes
х  All Typical Mistakes from BEM lesson theory are checked.
х Code follows all the [Code Style Rules]

anastasiiavorobiova (changes requested):
Great work! https://www.w3schools.com/css/css3_backgrounds.asp 
Screenshot 2024-08-18 at 15 08 37
The icon disappears on the focus. Consider fixing it 
You can use Pixel perfect extension to compare your solution with Figma and pass tests

VolodymyrV7:
the magnifying glass doesn't disappear, but I don't understand where the error is with the text size

lerastarynets (changes requested):
Pls don’t ask for help in github, because this way mentors are only able to see your questions once the review is requested. Instead, pls always ask for help in **fe_chat** if you don't understand something so mentors can quickly help you

VolodymyrV7:
ok)

src/style.css
```diff
@@ -1 +1,60 @@
...
+}
+
+.small-input {
+  flex: 1;
+  padding-left: 33px;
+  height: 42px;
+  font-size: 14px;
+  border-radius: 4px;
+  border: 1px solid #e1e7ed;
+  box-sizing: border-box;
```
TarasHoliuk, line 43: this property is better to define in * selector and remove everywhere else

src/style.css
```diff
@@ -1 +1,60 @@
...
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  background: #fff url('./images/Search.svg') no-repeat left 13px center;
+  background-size: 11px 11px;
+}
+
+input:hover {
```
TarasHoliuk, line 49: do not add styles using tag selector

Instead, add a common `input` class to both inputs and add all shared styles using this class (instead of repeating them in small-input and biginput)

src/style.css
```diff
@@ -1 +1,60 @@
...
+}
+
+input:focus {
+  background: linear-gradient(180deg, #fff 0%, #f6f6f7 100%)
+    url('./images/Search.svg') no-repeat left center;
+  font-weight: 900;
+  outline: none;
+  color: inherit;
+}
+git add readme.md
```
TarasHoliuk, line 60: shouldn't be here)

src/style.css
```diff
@@ -1 +1,60 @@
...
+  box-shadow: 0 1px 8px 0 rgba(61, 78, 97, 0.1);
+  background: #fff url('./images/Search.svg') no-repeat left 13px center;
+  background-size: 11px 11px;
+}
+
+input:hover {
+  box-shadow: 0 3px 8px 0 rgba(61, 78, 97, 0.2);
+}
+
+input:focus {
```
TarasHoliuk, line 53: one of the reason tests fail might be overwritten by standard browser styles for placeholder

Add correct styles (font family - inherit, color, font weight) using pseudoelement:
`.input::placeholder`

https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder

TarasHoliuk (changes requested):
Hi! GJ 👍 

I added suggestions to improve your code and, maybe, fix tests)

anastasiiavorobiova (approved):
Great work!

PR #4678, opened 2024-12-26
add task solution

Bukavyna:
https://github.com/Bukavyna/layout_search-bar-airbnb 
https://bukavyna.github.io/layout_search-bar-airbnb/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,22 +11,46 @@
       content="ie=edge"
     />
     <title>Document</title>
+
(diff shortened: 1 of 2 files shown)


volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. You need to fix width for inputs 
image

2. Fix styles like this and add `box-sizing: border-box` to inputs 
image

PR #4996, opened 2025-06-29
added input type of submit to index.html file

dewits15:


dewits15:
What is the issue with the checking feature?

PR #5061, opened 2025-08-05
add task solution

diman2807:
- [DEMO LINK](https://diman2807.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://diman2807.github.io/layout_search-bar-airbnb/report/html_report/)

### Requirements:
- use images from [src/images](src/images)
- there must be two search bars
- search bar must stretch the full width
- distance between two search bars must be 20px
- the big search bar must have top indent 20px
- follow styles from the mock
- default `font-weight` must be 300
- a search bar has 3 state default, `hover` and `focus`
- don't use JavaScript
- use `@font-face` for fonts
---

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -21,12 +21,27 @@ ___
 - add attribute `data-qa="big"` for big search form, and `data-qa="small"` for small
 - add attribute `data-qa="keypress"` to input in big search form
 ---
+
+Вимоги:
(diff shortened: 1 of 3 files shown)


vitalii-bondarenko-ua (approved):
Looks good.

PR #5094, opened 2025-08-25
Implement search bar task

VitaliiSamsonov:
- Implemented two search bars (big and small) according to mockup
- Added @font-face for Avenir font with weights 300 and 900
- Fixed margins between forms (20px)
- Added data-qa attributes: big, small, and keypress
- Updated styles for hover and focus states


VitaliiSamsonov:
@etojeDenys How do I do this? I don't understand.

PR #5244, opened 2025-11-30
Implement layout_search-bar-airbnb

Radek-Ptak:
- [DEMO LINK](https://Radek-Ptak.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Radek-Ptak.github.io/layout_search-bar-airbnb/report/html_report/)


natalia-klonowska (changes requested):
tests show that margin should be much smaller so remove padding from `main` element and keep the default body margin (remove `margin: 0` style from `body`)
image

natalia-klonowska (changes requested):
font and color aren't applied. Add `font-family` style to `search__input` class and use `::placeholder` pseudo-element to define color. also there is a slight difference in height so please specify exact height each input should have

natalia-klonowska (approved):
Great work! 🚀

PR #5378, opened 2026-03-26
finished doing additional task

ptrodion:
Icon implemented using background-image CSS property
 Inputs are written inside of 'form' tag with correctly passed attributes
 All Typical Mistakes from BEM lesson theory are checked.
 Code follows all the [Code Style Rules ❗️](https://github.com/ptrodion/layout_search-bar-airbnb/blob/master/checklist.md)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -25,8 +25,8 @@ ___
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_search-bar-airbnb/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_search-bar-airbnb/report/html_report/)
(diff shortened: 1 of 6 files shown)


brespect (changes requested):
Les pass all tests and provide the DEMP LINK to the PR description

ptrodion:
It fails the test because of the placeholder="Try 'Los Angeles'" in the bottom field. I've tried every possible way to add quotes, but the linter always changes their value, and the test fails. Even when I disabled the rule locally where the linter checks for double quotes and forced the text to match the test's requirements, the test still failed. This is very strange to me; could you please look into what the problem is? Because I've spent a lot of time on a problem that should be solved with a single line of code using &qout;TEXT&qout;

2pasha (approved):
lgtm 🔥

PR #5408, opened 2026-04-21
Develop

nyarachun:
- [ x ] Icon implemented using background-image CSS property
- [ x ] Inputs are written inside of 'form' tag with correctly passed attributes
- [ x ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ x ] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -23,10 +23,10 @@ ___
 ---
 ## Checklist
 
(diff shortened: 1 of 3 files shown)


brespect (changes requested):
Good progress, but you need to pass all CI checks before requesting the review

PR #5435, opened 2026-05-20
Add blocks search

Tryzub-web:
- [DEMO LINK](https://Tryzub-web.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Tryzub-web.github.io/layout_search-bar-airbnb/report/html_report/)


brespect (changes requested):
Good progress, but a bit different from the mockup, so try to pass all tests

Tryzub-web:
> Гарний прогрес, але трохи відрізняється від макета, тож спробуйте пройти всі тести

Tryzub-web:
> Гарний прогрес, але трохи відрізняється від макета, тож спробуйте пройти всі тести

Hi! I have an issue with a placeholder attribute that contains double quotes.

In the Figma mockup the text is: Try "Los Angeles"
The tests also expect: Try "Los Angeles"

I tried writing it as placeholder="Try &quot;Los Angeles&quot;" but after running npm run test, the linter automatically reformats it to placeholder='Try "Los Angeles"' — and then the same linter throws an error about inconsistent quote style in HTML attributes.

Also, I noticed that in the Figma mockup the small search input has no quotes in the placeholder, but the tests expect them to be there.

Could you advise what's the correct approach here?

Tryzub-web:
> Good progress, but a bit different from the mockup, so try to pass all tests

I tried changing the attribute quotes from double to single, but the linter throws this error:

error  The attribute "placeholder" is not double quoted  attr-quote-style

So if I use double quotes — the linter complains about the quotes inside the value.
If I use single quotes — the linter complains that the attribute is not double quoted.

I'm stuck and not sure how to pass the tests.

PR #5485, opened 2026-07-22
Airbnb search bar

HiljSV:
- [DEMO LINK](https://HiljSV.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://HiljSV.github.io/layout_search-bar-airbnb/report/html_report/)

- [x] Code follows all the Code Style Rules


HiljSV:
@mateacademy-ai-mentor Thanks for the review. I fixed the third point and would like to clarify the first two.
1. Fixed - '*' selector. Removed from both CSS files. 'box-sizing: border-box' and 'padding: 0' are now applied only to '.search-form__input', the single element that needs them. All 6 backstop scenarios still pass.
2. 20px gap between the search bars. The gap is already there: '.search-form { margin: 20px auto 0; }' is applied to both forms, so the second form has 'margin-top: 20px'. Arithmetic from the rendered page at 1024px: 8px (default body margin) + 20 + 70 (big form) + 20 (gap) + 42 (small form) = 160px, and the rendered document is exactly 1024x160 - identical to the reference screenshot generated from the official solution. Adding another 'margin-top' would make it 180px and fail all scenarios.
3. 'data-qa="hover"' on the small input. I did not add this attribute - it ships in the task's own starter src/index.html' (line 28), and it is also present in the official Mate Academy solution: https://mate-academy.github.io/layout_solutions/search-bar-airbnb/ . Removing it would make the markup differ from both. The hover state itself is implemented in CSS via ':hover', as you noted; the attribute is only a test hook.
Could you confirm?

HiljSV:
@mateacademy-ai-mentor What is the status of the check?

PR #5488, opened 2026-07-22
Implement Airbnb search bar layout

ArtMay777:
- [DEMO LINK](https://ArtMay777.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://ArtMay777.github.io/layout_search-bar-airbnb/report/html_report/)


ArtMay777:
@mateacademy-ai-mentor Thanks for the review. I believe all three points are already correct - could you re-check?

1. **20px gap.** Both forms share the base class `search-form`, which has `margin: 20px auto 0`. The `--big` / `--small` modifiers only set height, so the small form does have `margin-top: 20px`.

2. **`data-qa="keypress"` placement.** It is on the big form's input:
   `<form class="search-form search-form--big" data-qa="big">` → `<input class="search-form__input search-form__input--big" data-qa="keypress">`.
   The small form's input has `data-qa="hover"`.

3. **`data-qa="hover"` on the small input.** This attribute comes from the official starter template (`src/index.html`, line 28 in the original repo) — I kept it rather than added it.

All backstop scenarios pass and CI is green.

PR #5492, opened 2026-07-26
add task solution

vlad4uikov:


src/style.css
```diff
@@ -1 +1,89 @@
...
+  font-size: 16px;
+  font-weight: 300;
+
+  height: 100%;
+  width: 100%;
+
+  border: 1px solid #e1e7ed;
+  border-radius: 4px;
+
+  box-shadow: 0 1 8 0 #3d4e611a;
```
rostyslav-horytskyi, line 49: Invalid box-shadow syntax. Should be `box-shadow: 0 1px 8px 0 #3d4e611a;`. Without `px`, the shadow won't render at all. Same issue on line 59

src/style.css
```diff
@@ -1 +1,89 @@
...
+}
+
+.search__input:hover {
+  box-shadow: 0 3 8 0 #3d4e6133;
+}
+
+.search__input:focus {
+  background: linear-gradient(#fff, #f6f6f7);
+
+  text-shadow: 0 4 4 0 #00000040;
```
rostyslav-horytskyi, line 65: Should be `text-shadow: 0 4px 4px 0 #00000040;`. Also, `text-shadow` only accepts 2–3 length values (offset-x, offset-y, blur-radius), not 4. Remove the extra 0

rostyslav-horytskyi (changes requested):
Overall looks good, fix the errors and it'll be awesome!🔥

